PR #2, opened 2025-02-14
Add full set of ruff linting rules

jacobcook1995:
This PR tells the `ruff` linter to use a much broader set of code quality rules. The set that I've added is the full set used in the `virtual_ecosystem` repo. If this becomes too restrictive for the purposes of this repo we can deactivate certain rules


davidorme (approved):
LGTM

PR #3, opened 2025-02-14
Fixing R linting and fomatting in VSCode

jacobcook1995:
Updated the system dependencies so that the packages needed to get VSCode to play nicely are included


.vscode/settings_template.json
```diff
@@ -1,11 +1,11 @@
 {
     "r.alwaysUseActiveTerminal": true,
     "r.bracketedPaste": true,
-    "r.rpath.mac": "/opt/homebrew/bin/R",
+    "r.rpath.mac": "Insert output of `which R` (MacOS, Linux, WSL) or `where.exe R` (Windows Powershell)",
```
jacobcook1995, line 4: I'm guessing the option used is probably different if you aren't on MacOS (unless I'm misunderstanding  what `path.mac` means)

davidorme: Excellent point!

davidorme (approved):

PR #13, opened 2025-03-04
Templates for R file metadata

davidorme:
This PR:

* Provides templates for R script files and R Markdown containing the expected metadata fields as YAML formatted front matter. We aren't necessarily sticking with R Markdown as a notebook format, so might be better to start with R scripts, but we can also convert files easily enough down the line.

* TODO - add a simple checker to run validation of the metadata front matter.


annarallings:
Maybe I am looking at it incorrectly but it seems the templates are all ordered slightly differently. I am going to suggest we have one single yaml template to make peer review easy (it can be .R or even just txt). Those using markdowns can simply replace their header material with the text, as I understand it.

davidorme:
The problem @annarallings is that R (or python) scripts cannot have a random block of YAML at the start. It has to be commented in the way shown in that R file. So if we just provided the YAML file then everyone has to comment it out correctly. I think it is less error prone to provide versions for each one.

davidorme:
@annarallings @jacobcook1995 

These are my suggested templates - I think we should drop `R_version` (and delete it from these templates)  and I agree we should check the order is the same across the templates.

I really think there is value in having a version for each file type. I've added some other simple good practice stuff into the templates too. I've also created the start of a suite of maintenance tools that we can use to check the codebase and data.

Can you have a look?

jacobcook1995 (approved):
LGTM!

I also agree on dropping the version of R used, it's something that should be handled by dependancy management down the road

davidorme:
I've added a README to explain the templates a bit more and renamed the pure YAML file so that people don't think it is a template.

annarallings (approved):
LGTM

PR #15, opened 2025-03-05
Removing `renv`

davidorme:
This PR removes all of the files associated with using `renv`. It is troublesome to use and we don't want barriers in adoption right now.


jacobcook1995 (approved):
So long `renv` 👋

PR #30, opened 2025-03-23
adding a Jupyter notebook template

nickwctan:
I am currently working on reworking my script in Jupyter lab (notebook) so I thought it would be a good idea to create a template for it too. 

I might have terribly messed up the yaml metadata format as I have just learnt that yaml is very sensitive to tabs as indentation. Please advise.

### DIFF
--- a/templates/Jupyter_notebook_R_template.ipynb
+++ b/templates/Jupyter_notebook_R_template.ipynb
@@ -1,0 +1,152 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


davidorme:
@sphinxdrake It is actually much less clear how we store the metadata in Jupyter notebook format. The `.ipynb` format is a JSON document, so simply having the metadata in a text cell at the start doesn't work in the same way. We _can_ do this in the same way with a different format of the Jupyter notebook (MyST Markdown) but then we don't get the images in the same nice way.

Let's park this for the moment.

davidorme:
I'm closing this until we revisit Jupyter here.

PR #32, opened 2025-03-27
fix typo in anal(a)ysis folder name

hrlai:
This PR simply rename the "analaysis" folder to be "analysis" so both plant and soil scripts are in the same place, this will make it clearer for others, i.e, we only have one analysis folder (!)


davidorme (approved):
Was meaning to do this myself, since I put the typo in. Thanks for catching it.

arne-exe:
@hrlai thanks for this! Could you please briefly explain the process of how you did this?
I'm trying to better understand the process of changing existing folders in Git.

davidorme:
@arne-exe You need to use the `git mv` command. In this case @hrlai would have used `git mv analaysis analysis` - that moves all the files under `analaysis` to new paths under `analysis`.

And yes, because Git only cares about actual _file_ paths it does leave the empty directories behind. So you can now delete those folders.

arne-exe:
Great, thanks!

hrlai:
Yeap basically what @davidorme said 👆🏼 !

PR #40, opened 2025-04-07
Uploading ERA5 climate data python script

lelavathy:
This Python script retrieves ERA5-Land climate data from the Copernicus Climate Data Store (CDS). It automates the download climate data, which is useful for hydrological modeling, climate studies, and ecosystem analysis.
The script utilizes the `cdsapi` package to request ERA5-Land climate data for a specified year, covering all months and days.

package_dependencies:
    - cdsapi
    - netCDF4
    - pandas
    - xarray

 usage_notes: |
  - **Copernicus Data Store (CDS) Registration & API Key Setup**  
    To access ERA5-Land data, users must register for a free account at the [Copernicus Climate Data Store (CDS)](https://cds.climate.copernicus.eu/).  
    After registration, an API key is provided, which must be configured in a local `.cdsapirc` file for authentication.  
    The file should be placed in the user’s home directory (`~/.cdsapirc` on Linux/macOS or `C:\Users\YourUsername\.cdsapirc` on Windows) and contain the following format:  

    ```ini
    url: https://cds.climate.copernicus.eu/api/v2
    key: your-uid:your-api-key
    verify: 1
    ```

  - **Installing and Configuring the `cdsapi` Package**  
    The `cdsapi` Python package is required to communicate with the Copernicus Data Store. Install it using:  
    ```bash
    pip install cdsapi
    ```
    Ensure the `cdsapi.Client()` function is correctly initialized in the script before making any requests.

  - **Download Time Considerations**  
    The dataset request spans an entire year (12 months, 365 days, 24-hourly data points per day).  
    Due to the large data volume, download times vary based on:  
      - Internet speed  
      - CDS server response time (which may slow down during high demand)  
      - The selected spatial extent (smaller regions will download faster)


vgro (dismissed):
This looks good to me, a few minor comments. Make sure that you run `pre-commit` until all the tests pass before you merge :-)

vgro (changes requested):
Tests need to pass :-)

analysis/abiotic/ERA5-Land Rainfall Data Download Script-General.py
```diff
@@ -0,0 +1,96 @@
+#' ---
+#'title: ERA5-Land Rainfall Data Download Script-General
+
+#'description: |
+#'   This Python script retrieves ERA5-Land rainfall data from the Copernicus Climate Data Store (CDS).
+#'   It automates the download of high-resolution precipitation data, which is useful for hydrological
```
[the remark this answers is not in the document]
lelavathy: instead of high-resolution, i can change to hourly rainfall data

analysis/abiotic/ERA5-Land Rainfall Data Download Script-General.py
```diff
@@ -0,0 +1,96 @@
...
+    "month": [f"{i:02d}" for i in range(1, 13)],  # January to December
+    "day": [f"{i:02d}" for i in range(1, 32)],  # Days 1 to 31
+    "time": [f"{i:02d}:00" for i in range(24)],  # Hourly data (00:00 to 23:00)
+    "data_format": "netcdf",
+    "area": [
+        N,
+        W,
+        S,
+        E,
+    ],  # specify your own bounding box coordinates (North, West, South, East)
```
davidorme, line 87: These values aren't defined anywhere in the file. It makes sense to allow this script to be used with different locations, but we have to specify what data we are requesting. So either we put values here for now or we make it into a command line tool that accepts arguments and have a separate bash script to get the SAFE regional data (or whatever it is we want 😄 ). I think fix it for now.

analysis/abiotic/ERA5-Land Rainfall Data Download Script-General.py
```diff
@@ -0,0 +1,96 @@
...
+# Initialize the CDS API client
+c = cdsapi.Client()
+
+# Define dataset and request parameters
+dataset = "reanalysis-era5-land"
+request = {
+    "variable": ["total_precipitation"],
+    "year": "<year>",  # Replace with your desired year, e.g., "2020"
+    "month": [f"{i:02d}" for i in range(1, 13)],  # January to December
+    "day": [f"{i:02d}" for i in range(1, 32)],  # Days 1 to 31
```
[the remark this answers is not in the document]
davidorme: I believe we can set this as a range. The example on the `cdsapi` repo shows a date range:

https://github.com/ecmwf/cdsapi?tab=readme-ov-file

We also need to be able to set the year programatically, not by having to edit and rerun this.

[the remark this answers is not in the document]
lelavathy: CDS API is designed to automatically ignore those non-existent days. When it receives requests for invalid dates (like June 31), the server just skips them, so you won't get errors. So, you won't get errors

davidorme (changes requested):
This is great, but as it stands we can't recreate the data download because this has a bunch of placeholders that need to be manually edited. We need to turn this into a tool where we can document the usage.

If we made it into a command line tool, it would have a _lot_ of arguments (bbox, region name, file name, year). We could add them using `argparse` but could also use e.g. a TOML file of the requested details to have a single input?

analysis/abiotic/ERA5-Land Rainfall Data Download Script-General.py
lelavathy: You're absolutely right — the current script is set up as a general template, but I agree it’s best to fix it for now to ensure clarity and reproducibility. I’ll go ahead and set the script to download total precipitation data for the Maliau Basin region, with a defined year range from 2010 to 2020.

analysis/abiotic/ERA5-Land Rainfall Data Download Script-General.py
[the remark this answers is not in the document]
vgro: that's perfect. There is the spatial and temporal component and this needs to be clear :-)

analysis/abiotic/ERA5-Land Rainfall Data Download Script-General.py
davidorme: Will we use the same API to get other variables?

davidorme: Oh - one other thing. It is best to avoid spaces in file names - it can cause problems with command line tools, so this would be better as `ERA5-Land_Rainfall_Data_Download_Script-General.py` or something like that.

lelavathy: I don’t think we’ll be using the same API for other variables — at least for now — because I’m currently using my personal API key to download the ERA5 data. We might need to discuss setting up a shared or project-specific API access if we plan to expand the scope to include more variables or allow others to run the script independently.

davidorme: We'd have to share those credentials outside of the repository anyway. One thing you shouldn't add to a repo are any kind of API keys and credentials.

annarallings: Agree with David - Let's put in the Maliau example and we can use this script to store some use cases. The goal being to make our data collection reproducible. Future users can always select the scenario they want in the script run, but it's good to track exactly what the configuration of each scenario is.

lelavathy: I’ve created two scripts to address the issues. 
1. Script with argparse function: This script is designed to handle different variables and years via command-line arguments. It allows the user to download a specific variable for a given year, region, and bounding box. It's flexible and can be reused for multiple locations or variable requests. **(refer to script: era5_land_climate_data_download.py).** It’s recommended to download one variable per year at a time to avoid overwhelming the system. This is why I created the argparse script, as it allows me to download multiple variables for different years and locations in separate requests.

2.Script for Maliau Basin: This script is tailored specifically for downloading precipitation data from the Maliau Basin for the years 2010-2020. However, I ran into an issue where it causes an error due to the limitations on downloading large datasets from ERA5. The problem is that the data volume exceeds what the server can handle in one request.**(refer to script: era5_rainfall_data_download_maliau.py.** To address the problem of large data volumes, I can modify the script with argparse function to allow the user to specify a single year per download. This way, the script will download data for one year at a time, avoiding issues with the data size limitations.

At the moment, I’m just using the script with the argparse function, as I prefer to use one script for all climate variables. In this setup, I can run the script from the command line arrguments to adjust the coordinates, location, variable, and year for each download. Which script would be the best approach for the repository?

vgro: I think it would be good to keep both the generalisable and the Maliau specific script - one can be adopted for any case and one is specific to our experiment.

One thing I just realized when I read about the limitations is that we do not need hourly data for our April simulations, this would be interesting at some later stage but monthly is what the model runs with now. So you should be able to download all the necessary variables in one go in monthly resolution, there is an ERA5 -land monthly dataset on the CDS.

davidorme:
@lelavathy Did you intent to close this PR?

lelavathy:
Actually i didnt intent to close the pr, just thought to change rainfall to
climate for more general Era5 data download

lelavathy:
@Orme, ***@***.***>Thanks for checking, David! I didn’t mean to close the PR. After renaming the branch, GitHub prompted me to create a new PR, but I didn’t realize it would ask for that. I’m still figuring things out and am in the process of learning Git and GitHub effectively. Apologies for the mix-up! I’ll make sure to keep everything in the same PR going forward by adding new commits. I really appreciate your understanding and patience!
________________________________

PR #52, opened 2025-05-16
Updated PFT cohort distribution script

### LINKED ISSUE #49: Update PFT cohort distribution script
This issue should update the PFT cohort distribution script so that the output file provides a count of individuals, instead of a row for each unique individual.
It should also express the cohort distribution on a per area basis, instead of just "X individuals across the 3 old growth plots".
To do this, define what area should be used (e.g., the area of a cell used in the model, or any other standardised/reproducible basis such as per hectare for example).

arne-exe:
This PR fixes #49 by:

- Updating the script
- Updating the output file
- Uploading an additional output file

(Note to self: also update the output files in Globus once finished)


arne-exe:
Not really sure what went wrong in this step here

davidorme:
> Not really sure why it did this or if this is the correct way of amending a commit message.

Once you've pushed a commit, that's pretty much it, it's in the record [^1]. You need to push a new commit to get it to how it should be. If you _haven't_ pushed the commit then you can use `git commit --amend` to change the commit message itself or you can use `git reset` to move your clone back to before the commit and then do it again right. 

[^1]: That's technically not true. You can change the repo history on GitHub, but it's typically a very very bad idea.

davidorme (approved):
LGTM.

I think you're overusing `nolint` a bit but that's just aesthetics and the "correctly" formatted code can often jar. I used to get this a lot with Python and `ruff` but in the end I just let the autoformatting do what it wants. Sometimes it makes me squint a little, but its quicker to let it do it 😄

arne-exe:
> LGTM.
> 
> I think you're overusing `nolint` a bit but that's just aesthetics and the "correctly" formatted code can often jar. I used to get this a lot with Python and `ruff` but in the end I just let the autoformatting do what it wants. Sometimes it makes me squint a little, but its quicker to let it do it 😄

Yeah, it's bringing up error messages more often than I'd like, and I don't always see a better way to format it. I think at the moment it mostly has issues with variable names or certain small parts of commented code. If it's just aesthetic changes then I prefer not to spend too much extra time on trying to "fix" them.
When you refer to autoformatting, this does not include the lines where I used "nolint" right? Or do you mean that there is an autoformatting option that can help solve these?

davidorme:
> Yeah, it's bringing up error messages more often than I'd like, and I don't always see a better way to format it. I think at the moment it mostly has issues with variable names or certain small parts of commented code. If it's just aesthetic changes then I prefer not to spend too much extra time on trying to "fix" them. When you refer to autoformatting, this does not include the lines where I used "nolint" right? Or do you mean that there is an autoformatting option that can help solve these?

Are you using VSCode or RStudio? In VSCode I get the offending lines highlighted in blue and hover text tells me what the issue is. My guess is that 95% of your issues are either:

* Using "overlong" variable names. I'd call these descriptive variable names and we can update the rules to make this go away.
* Using "non-standard" variable names. Like `OG1_plots` has [gasps] _capitals_ in it. It wants `og1_plots`. Again we can update the rules to make it shut up about this.
* Over long lines - often hard to fix if you have long variable names 😄 or long path names.

So, we can review what we want to be a problem. 

@annarallings - can you also keep an eye on what people are using `# nolint` on and then we can finesse the rules for our code base and stop the linters moaning about stuff we don't care about.

arne-exe:
> 
> Are you using VSCode or RStudio? In VSCode I get the offending lines highlighted in blue and hover text tells me what the issue is. My guess is that 95% of your issues are either:
> 

I'm using both, VSC indeed has that useful functionality but I've noticed that it sometimes fails to do the autoformatting. For example, when you separate a line of code in RStudio, it will automatically apply the correct spacing, whereas in VSC it often fails to do this correctly. Even worse, sometimes VSC says a spacing of X is required, which I then apply, but then it still says it's wrong. I believe the reason for this is that the first "enter" is not in the correct location, which then messes up everything else. So VSC knows that something is wrong with the spacing, but it doesn't get why, and provides the incorrect solution. 
To handle this, I find it to be the easiest to handle splitting up lines of code and spacing in Rstudio while starting from the particular code as 1 "long" line, and then separate it into smaller sections while RStudio correctly does the spacing for me. I will check again if VSC can do the spacing when starting from 1 "long" line as well. 

> * Using "overlong" variable names. I'd call these descriptive variable names and we can update the rules to make this go away.
> * Using "non-standard" variable names. Like `OG1_plots` has [gasps] _capitals_ in it. It wants `og1_plots`. Again we can update the rules to make it shut up about this.
> * Over long lines - often hard to fix if you have long variable names 😄 or long path names.
> 

I agree, those are the most frequent issues I encounter!

Issue #51, opened 2025-05-15
Create script allocation to reproductive tissue mass and propagule mass


arne-exe:
This issue should create and upload a script that calculates:
- the ratio that allows to derive reproductive tissue carbon mass from foliage carbon mass
- the ratio that allows to derive propagule carbon mass from reproductive tissue carbon mass

PR #53, opened 2025-05-21
Updated plant stoichiometry script

### LINKED ISSUE #48: Update plant stoichiometry script
This issue should update the plant stoichiometry script by calculating and adding:

- reproductive tissue stoichiometry
- lignin content for leaves, wood, roots, reproductive tissues
- carbon mass per propagule

Preferably these values are at PFT level, for different tissue types (like senesced / non-senesced leaves, and sapwood / heartwood, etc.).
This won't always be possible, so in some cases a generic value for dipterocarp forests will be used instead.

arne-exe:
This PR fixes #48 by:

- Adding reproductive tissue stoichiometry (including fruits/seeds and flowers)
- Adding lignin content
- Updating the output file (removing some unused variables and adding new ones)

(Note to self: also update the output files in Globus once finished)

Still to do:
- ~~Add meta data sources reproductive tissue data~~
- ~~Add fine root stoichiometry~~
- Add lignin content for all tissue types (~~leaf, wood, root~~, ~~fruit/seed~~, flower)
- ~~Add carbon_mass_per_propagule~~


arne-exe:
@sallymatson or @davidorme, regarding tissue lignin content: this is typically given as a percentage of the tissue dry weight. As far as I'm aware the model mostly works with tissue carbon mass, and not dry weight. Do you have a preference as to how tissue lignin content is prepared (i.e., percentage of dry mass or percentage of carbon mass)?

arne-exe:
> @sallymatson or @davidorme, regarding tissue lignin content: this is typically given as a percentage of the tissue dry weight. As far as I'm aware the model mostly works with tissue carbon mass, and not dry weight. Do you have a preference as to how tissue lignin content is prepared (i.e., percentage of dry mass or percentage of carbon mass)?

For leaf lignin content I've calculated it as the percentage of leaf carbon mass, where the resulting amount of lignin will have the same unit as leaf carbon mass (e.g., kg C). I think this should make things easier to work with in the model compared to expressing lignin content on a dry weight basis. Lignin is mostly made up of carbon, with percentages of N and P being close to zero. If the main objective for lignin content's use is to access that pool of carbon then I think this approach is fine and does not need further change. If the model needs to quantify all other nutrients then some new functionality will probably need to be added in the model (like obtaining dry weight lignin from foliage carbon mass, and then separating dry weight lignin into its respective nutrients including/other than carbon, i.e. using some lignin stoichiometric ratios).

arne-exe:
When reproductive tissue carbon mass is derived from foliage carbon mass, and when non-propagule carbon mass is then derived from reproductive tissue carbon mass, does this take place on a "most limiting nutrient" basis? For example, will the size of that respective biomass pool be constrained by the nutrient that is limited first (taking the stoichiometric ratios into account)? If so, what happens to the rest of the (non-limited) nutrients in that biomass pool?

davidorme:
> For leaf lignin content I've calculated it as the percentage of leaf carbon mass, where the resulting amount of lignin will have the same unit as leaf carbon mass (e.g., kg C).

I think this is correct. The main reason to track lignin is that it is handled differently in the decay pools, so we really just need to know how much of the leaf goes into which pool. @jacobcook1995 You don't need anything other than carbon mass here, right?

davidorme:
> When reproductive tissue carbon mass is derived from foliage carbon mass, and when non-propagule carbon mass is then derived from reproductive tissue carbon mass, does this take place on a "most limiting nutrient" basis? For example, will the size of that respective biomass pool be constrained by the nutrient that is limited first (taking the stoichiometric ratios into account)? If so, what happens to the rest of the (non-limited) nutrients in that biomass pool?

At this point, the size of the pools are fixed. The nutrient content will vary depending on availability, so the stochiometric ratios aren't fixed.

jacobcook1995:
> > For leaf lignin content I've calculated it as the percentage of leaf carbon mass, where the resulting amount of lignin will have the same unit as leaf carbon mass (e.g., kg C).
> 
> I think this is correct. The main reason to track lignin is that it is handled differently in the decay pools, so we really just need to know how much of the leaf goes into which pool. @jacobcook1995 You don't need anything other than carbon mass here, right?

Yeah all I need is a the proportion of total carbon mass that is lignin. So lignin as a mass works fine (as total carbon is tracked anyway)!

arne-exe:
> > When reproductive tissue carbon mass is derived from foliage carbon mass, and when non-propagule carbon mass is then derived from reproductive tissue carbon mass, does this take place on a "most limiting nutrient" basis? For example, will the size of that respective biomass pool be constrained by the nutrient that is limited first (taking the stoichiometric ratios into account)? If so, what happens to the rest of the (non-limited) nutrients in that biomass pool?
> 
> At this point, the size of the pools are fixed. The nutrient content will vary depending on availability, so the stochiometric ratios aren't fixed.

Maybe I'm doing a poor job at explaining my thoughts here, I'll try again:
- Certain pools are derived from others (reproductive tissue mass as a percentage of foliage mass), and this works through carbon mass
- The carbon mass of the reproductive tissue pool is then used to calculate N and P mass of that pool
- Non-propagule carbon mass is then derived from reproductive tissue carbon mass using a percentage
- This non-propagule carbon mass is then used to calculate N and P mass of that pool
- The same is done for propagule C, N and P mass
- Importantly, the N and P mass of non-propagule and propagule pools combined do not necessarily amount to the respective nutrient pools of the reproductive tissue pool (as they are not derived from the same source/dataset).

Whether this discrepancy (i.e., nutrient pools not adding up) is an issue (or if it exists in the first place) I don't know, but I thought it was worth pointing out

arne-exe:
> 
> Yeah all I need is a the proportion of total carbon mass that is lignin. So lignin as a mass works fine (as total carbon is tracked anyway)!
>

Alright, I did correct the lignin content percentage for lignin carbon content (62.5%) but the final percentage should give you the amount of carbon that is present in lignin (just checking that this correction is not already in the model to avoid correcting for it twice)

arne-exe:
> To represent carbon_mass_per_propagule I've added mature_fruit_C_mass. This is based on only 1 dipterocarp species, and uses the dry weight and carbon content of mature fruits. In the future, I'd like to refine this value to be PFT specific, and down the line this should also be updated to represent the seed mass instead of whole fruit mass.

I found another study that had seed mass data for that same species (+10 other species) so in the script I also calculated seed_C_mass (although it still uses the stoichiometry of the mature fruit as data for seed stoichiometry was not available). 

In general, studies/data on fruit and seed stoichiometry are rather scarce, so I think it will be hard to get PFT specific values for this.

At the moment, the reproductive tissue pool can be separated into A) flowers and B) fruits and seeds. This separation will be done based on data from 1 dipterocarp species (Dipterocarpus tempehes), which I'll write out in issue #51. The fruit/seed pool is currently entirely made up of seeds where the total pool C mass divided by the carbon_mass_per_propagule to give the amount of seeds. Nitrogen and phosphorus in this pool can be calculated using the ratios provided, which are based on the mature fruit stoichiometry of that same dipterocarp species. Seed lignin content (as a percentage of seed carbon mass) is also based on data from that same species.

A few other things to note:
- No data on flower lignin content yet
- I think the model will currently assign all reproductive tissue carbon mass into seeds directly (i.e., it does not convert it into fruits, which are then separated into part seed and part other tissues). This means that the amount of seeds will be overestimated (as some of that carbon will in theory be used to construct parts of the fruit that are not seed).
- The variable "mature_fruit_C_mass" in the output file currently does not have a use.
- Lastly, given that it's pretty hard to find PFT specific data for fruit and seed stoichiometry, it might be worth discussing what would be the best approach to define "general/representative" values. For the example of seed lignin content, if PFT specific values are not available, would it be better to use the average ratio across 11 dipterocarp species, or would it be better to keep the current value as it "preserves" the relationships between variables within that species?

annarallings (approved):
LGTM

PR #58, opened 2025-05-30
Updated T model parameters script

### LINKED ISSUE #50: Update T model parameters script
This issue should update the T model parameters script by calculating and adding values for:

- ratio of fine root mass to foliage area
- turnover times for leaves, fine roots and reproductive tissue
- light extinction coefficient
- yield factor
- specific respiration rates for leaves, fine roots and reproductive tissue
- leaf area index

arne-exe:
This PR fixes #50 by:

- Adding additional T model parameters
- Updating the output summary file in the repo

To do:
- ~~ratio of fine root mass to foliage area (also added second approach using PFT specific SLA values)~~
- ~~turnover times for leaves, fine roots and reproductive tissue~~
- ~~light extinction coefficient~~
- ~~yield factor~~
- ~~specific respiration rates for leaves, stem and fine roots~~
- ~~leaf area index~~
(- crown gap fraction)
- Update output summary file

(Also add output to Globus)


arne-exe:
I can't figure out how Li et al. (2014) or Yan and Zhao (2007; paper referenced in Li et al.) obtained their values for "specific respiration rate". The closest we'll get using their values is a generic value for a "rain forest tree". There are detailed studies available for SAFE on the respiration of leaves, sapwood and fine roots, so if we can figure out what "specific respiration" refers to in the T model or how it is obtained, then this would help the calibration. For now I've used the same reference as Li et al, but using the value for rain forest tree instead of temperate coniferous tree.

annarallings (approved):
LGTM

PR #59, opened 2025-05-31
Model fungal-to-bacterial ratio

### LINKED ISSUE #57: Model fungal-bacterial ratio

hrlai:
This PR estimates fungal-to-bacterial ratio using the SAFE dataset [here](https://zenodo.org/records/3929632).

Main points:
- There are various soil covariates but I ended up included only soil pH and moisture, because the soil nutrients (N, P and C) are highly collinear with soil moisture. 
- I did not model ratio, instead I modelled fungal and bacterial biomass (in terms of PLFA) first, and then predicted their ratio. This will future proof our analysis in case we want to model OTUs in a joint-species-distribution-model framework (one day we will have the OTU data available)
- The fungal-to-bacterial ratio is estimated to be 0.24 --- this value stay roughly the same regardless of my modelling decision.

### DIFF
--- a/analysis/soil/fungal_bacteria_ratio/model.R
+++ b/analysis/soil/fungal_bacteria_ratio/model.R
@@ -40,6 +40,7 @@
 library(tidyverse)
 library(readxl)
 library(glmmTMB)
+library(performance)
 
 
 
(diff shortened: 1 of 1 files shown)


jacobcook1995:
> Might be worth giving this a quick looking before concluding our quick analysis here.
> 
> https://essd.copernicus.org/articles/14/4339/2022/

Eye balling that paper it looks like your number is basically consistent. It's interesting that that map seems to suggest that fungal ratios in the Maritime South East Asia are unusually high for the tropics, but there aren't any sampling points from the region included

jacobcook1995 (approved):
The scientific content all looks good!

I guess it would be good to add someone from the data team as a code reviewer, as my ability to give feedback on R code is pretty limited 🫠

davidorme:
> I guess it would be good to add someone from the data team as a code reviewer, as my ability to give feedback on R code is pretty limited 🫠

Look good to me. I don't personally use the pipe operator (`%>%`) but the usage here is well commented and it is for relatively short blocks of data handling that are easy to track.

PR #64, opened 2025-06-04
61 model mycorrhizal types

### LINKED ISSUE #61: Model mycorrhizal types
Currently the literature is full of AM _or_ EM compositions, but rarely both. I am going to start from a promising [dataset ](https://springernature.figshare.com/articles/dataset/A_curated_soil_fungal_dataset_to_advance_fungal_ecology_and_conservation_research_in_Australia_and_Antarctica/27938037)from Australia and limiting to their tropical sites, and this [global dataset](https://www.nature.com/articles/s41597-020-0567-7).

hrlai:
This PR uses the script `analysis/soil/mycorrhizae/model_SAFE.R` estimate the (relative) abundance of a few fungal guilds from the SAFE dataset https://doi.org/10.5281/zenodo.13122106 :
- arbuscular mycorrhizal
- ectomycorrhizal
- saprotroph
- pathogen
- parasite
- endophyte
- lichenized
- epiphyte
- The remaining lumped as "others"

Due to the use of an offset term, the model by default controls for total reads per sample. This means that the modelled / predicted abundance is expected count per _one_ read, i.e., relative abundance. That said, we can feed in different number of total reads to predict expected abundance per N total counts. 

In short, if we do nothing when predicting outputs / interpreting the intercepts, then the model is about relative abundances. 

NB: there is another script `model_global.R`, which tries to achieve the same goal using another global dataset, but I have decided to work on the SAFE data first, and leave this global analysis as WIP to return to later.


jacobcook1995 (approved):
This looks good to me!

I guess knowing the relative abundances we could estimate the absolute abundances based on data on total fungal biomass (does that sound like a sensible approach to you?)

(as always the caveat that I can't really give code style/quality feedback on R code)

PR #72, opened 2025-07-12
Disable object_name_linter

### LINKED ISSUE #71: `lintr` object_name_linter is not compatible with model coding
Hi @davidorme , would you consider relaxing `lintr` particularly the `object_name_linter` option for coding our models?

There has been multiple occasions where I really hope we do not need to strictly follow the snake_case criterion. 

For example, the following coding was not allowed:
```
C <- chem$C_perc / 100
^
log_mu_C <- normal(0, 2.5)
^~~~~~~~
sigma_C <- exponential(1)
^~~~~~~
sP <- exp(log_sP)
^~
structural <- (1 - fm) * exp(-r * L - ks * t)
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
but I think they are perfectly readable, and perhaps even preferable because these codes approximate how we write down the mathematical equations. But `lintr` doesn't like maybe because it thinks I'm mixing in camelCases ?

hrlai:
This PR disables the `object_name_linter` option in `lintr` to relax the object name conventions --- we'll just need to make sure each of us are consistent from hereon!

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,5 +1,6 @@
 linters: linters_with_defaults(
     cyclocomp_linter(complexity_limit = 10L),
     indentation_linter(indent = 2L),
-    line_length_linter(length = 88L)
+    line_length_linter(length = 88L),
+    object_name_linter = NULL
   )


davidorme:
@hrlai Why was this merged - it wasn't completed.

PR #75, opened 2025-07-13
74 generate netcdf soil

hrlai:
The output from this PR is a netCDF file that mimics the soil example data. `ve_example` generates the example data from python scripts, but here I mimic it in R. I tested it by replacing the python output with the R output, and successfully ran it in `ve_run`. 

The goal here is to simulate the scenario of an user, who will probably be an ecologist collecting data from the field and be experienced with R, but not python. 

From here a few things are very clear:
1. One would simply need to arrange the field data (or any input data) in a data.frame with these columns:
     - A longitudinal or "x" column
     - A latitudinal or "y" column
     - followed by additional columns, each being a biotic or abiotic variable used in VE
2. Next, convert the data.frame to arrays
3. Add whatever attributes to the arrays (units, dimension etc.)
4. Put all those variable arrays into a netCDF file and save the output
5. Either replace the netCDF in VE, or follow steps [here](https://virtual-ecosystem.readthedocs.io/en/latest/using_the_ve/data/data.html#loading-data-from-a-configuration) (which needs clearer documentation I feel, see my request [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/945))

I think this will be applicable to any other module that takes the same input (i.e., just netCDF inputs) as the soil module.


arne-exe:
Would it be a good idea to structure this part of the repo similar to the folder structure in the VE_example? Just to keep things structured and to make it easier for us to see how things link together? We could re-use the **config** and **generation_scripts** folder names and then we can also use this repo to prepare/discuss the configuration files. The data folder from the VE_example can then be the data folder from the VE_data_science repo. Any thoughts?

PR #81, opened 2025-07-16
80 Implement cohort distribution in plant generation script

### LINKED ISSUE #80: Implement cohort distribution in plant generation script
This sub-issue should implement the usage of the new cohort distribution in the plant generation script.
It will:
-Add a new folder "generation_scripts" similar to the VE_example
-Within this folder add a "plant" folder
-Copy the plant generation script from the VE example but update the cohort distribution section so that it uses the cohort distribution from the derived data folder in the VE_data_science repo
-Also copy the common.py script for now (keep within the plant folder, to allow plant-specific testing). We'll need to check later on how we handle the common.py script across models (see discussion in https://github.com/ImperialCollegeLondon/ve_data_science/discussions/66#discussioncomment-13741402)

To streamline the steps above, also change the cohort distribution script in the analysis folder so that:
-there are no more trees with "unknown" PFT (can revisit this idea again if needed)
-the cohort distribution is still standardized per hectare, but is not rounded up or down yet (as the plant generation script will perform scaling of the cohort distribution depending on the grid area specified in the common.py script). So, need to undo the rounding in the analysis script to prevent rounding twice.

arne-exe:
This PR should implement the usage of the new cohort distribution in the plant generation script.
It will:
-Add a new folder "generation_scripts" similar to the VE_example
-Within this folder add a "plant" folder
-Copy the plant generation script from the VE example but update the cohort distribution section so that it uses the cohort distribution from the derived data folder in the VE_data_science repo
-Also copy the common.py script for now (keep within the plant folder, to allow plant-specific testing).

To streamline the steps above, also change the cohort distribution script in the analysis folder so that:
-there are no more trees with "unknown" PFT (can revisit this idea again if needed)
-the cohort distribution is still standardized per hectare, but is not rounded up or down yet (as the plant generation script will perform scaling of the cohort distribution depending on the grid area specified in the common.py script). So, need to undo the rounding in the analysis script to prevent rounding twice.


generation_scripts/plant/plant_data.py
```diff
@@ -0,0 +1,68 @@
+"""Script to generate data to initialise the plants model.
+
+This script exports a NetCDF file containing the plant community setup for the 9 by
+9 example grid. Each cell contains 4 different plant functional types and 34 cohorts.
+
+"""
+
+import numpy as np
+import pandas as pd
+from virtual_ecosystem.example_data.generation_scripts.common import (
```
arne-exe, line 10: This needs to be replaced by ve_data_science.generation_scripts.plant.common and the common.py script also needs to be added (for now added to plant folder to allow plant-specific testing).

generation_scripts/plant/plant_data.py
```diff
@@ -0,0 +1,68 @@
...
+    n_dates,
+    time,
+    time_index,
+)
+from xarray import DataArray, Dataset
+
+data = Dataset()
+
+# Load cohort distribution
+plant_cohorts = pd.read_csv("../data/plant_cohorts.csv")
```
arne-exe, line 22: Need to recheck that relative paths match the repo, also need to add the option for absolute paths for "external" users.

generation_scripts/plant/plant_data.py
```diff
@@ -0,0 +1,68 @@
...
+# Spatio-temporal data
+data["downward_shortwave_radiation"] = DataArray(
+    data=np.full((n_cells, n_dates), fill_value=2040),
+    coords={"cell_id": cell_id, "time_index": time_index},
+)
+
+
+data["time"] = DataArray(time, coords={"time_index": time_index})
+
+data.to_netcdf("../data/plant_data.nc")
```
arne-exe, line 68: Also check paths here

generation_scripts/plant/plant_data.py
```diff
@@ -0,0 +1,68 @@
...
+)
+
+# Spatio-temporal data
+data["downward_shortwave_radiation"] = DataArray(
+    data=np.full((n_cells, n_dates), fill_value=2040),
+    coords={"cell_id": cell_id, "time_index": time_index},
+)
+
+
+data["time"] = DataArray(time, coords={"time_index": time_index})
```
arne-exe, line 66: >>> data["time"] = DataArray(time, coords={"time_index": time_index})
<stdin>:1: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values 
to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.

davidorme: Yeah - this is a pain in the arse.

There is a trade off in `numpy.datetime64` between the temporal resolution of a datetime and how large a timespan can be represented [https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-units]. We'd really probably only want to work with second precision, but `xarray` and `pandas` currently unavoidable do this. 

This _might_ come back to bite us because, using nanosecond precision, we only get a useable range of ~300 years. But hopefully the problem will go away before we run into this.

generation_scripts/plant/plant_data.py
arne-exe: I still have a lot to learn on how writing scripts in Python works.
It didn't find the working directory of the script in interactive mode (which I prefer to run the code line by line), I had to set it manually using an absolute path, to do this I had to import os and then specify the path but then there was an issue because not all the imports were at the top of the script (I had to define the working directory before being able to read in from the common script but this resulted in an error during the commit), so eventually I just had to delete setting the absolute path in order to get the commit through. 
Also, my pre-commit broke, not sure why (maybe because I was trying to commit both an R and Python script at the same time?). 
Also, sometimes it would accept my command to create the NCDF file but most of the time it said something like "NCDF4 not installed", also not sure why.

generation_scripts/plant/plant_data.py
davidorme: Honestly, I think it sounds like we're currently trying to do two things in the repo: 

1) get the data sorted for our use case
2) create a framework for future users to adopt

In my opinion, we should **not even remotely** be trying to do (2) because we don't know what that looks like yet and it adds complexity to doing (1).  I think it's a very useful thing to keep notes about and keep thinking about but we should just keep this simple.

generation_scripts/plant/plant_data.py
davidorme: No - absolutely raise this stuff. If something happens and it bothers you and we haven't responded then kick us till we do. Even it our response is - yeah that happens - then it's something you check off your list of things to worry about.

generation_scripts/plant/plant_data.py
arne-exe: > @arne-exe
> 
> I don't understand the issue here that requires absolute paths. Why is this needed?

A lot of this is trial and error for me @davidorme, plus it's pretty hard to do this in a new coding environment, so please keep that in mind. Worst case we just delete this entire generation folder if things turn out like a disaster

davidorme: > so please keep that in mind.

Oh I do - absolutely. 

And I also keep in mind that the whole team are also thinking much bigger picture than the immediate job and think about how this whole system might get picked up by the next users, which is fantastic. I just think those are two separate tasks.

generation_scripts/plant/plant_data.py
arne-exe: The original comment about absolute paths was before we discussed those things in the data input workflow discussion yesterday, and since this PR was marked as a draft I added those comments as temporary notes for myself.  

> Honestly, I think it sounds like we're currently trying to do two things in the repo:
> 
> 1. get the data sorted for our use case
> 2. create a framework for future users to adopt
> 
> In my opinion, we should **not even remotely** be trying to do (2) because we don't know what that looks like yet and it adds complexity to doing (1). I think it's a very useful thing to keep notes about and keep thinking about but we should just keep this simple.

I think you're right, I was trying to combine both in an attempt to be efficient, but as you mentioned it just makes the first step more complicated. So for now I'll just focus on getting the data sorted for our use case.

generation_scripts/plant/plant_data.py
arne-exe: > Actually - just checked in `xarray` and this has now been sorted (to some extent but good enough for our purposes!). If you update `xarray` then it should go away.

To what version did you update it? There seem to be some restrictions / compatibility issues with the VE and Poetry when I update to the newest version. I won't mess around too much with this, but it seems like this should then be updated in Poetry of the repo? Looking into this also made me realize that up until now I never really checked that I was working in the correct environment. I will have another look at the repo documentation you provided regarding Poetry, but I can imagine that the other members of the data_team may have also overlooked this. Perhaps worth mentioning this during one of the next meetings.

arne-exe:
@davidorme, I resolved the conflicts but I have a question. For the changes I wanted to accept from main I changed the script and pushed these changes as new commits. However, for the changes I didn't want to accept I removed the undesired option in the web editor, then committed these changes directly on GitHub. I then checked the status of my local branch, which was up to date (as I picked the option from my branch in the web editor).

My question now is what I would have to do in case I selected the other (main) option in the web editor. How would I then get my local branch up to date? Would that be by pulling from origin, and would than import the additional changes from the commit that was done directly on GitHub?

annarallings (approved):
LGTM - so sorry for the delay I don't know how I missed this!!

davidorme:
> @davidorme, I resolved the conflicts but I have a question. For the changes I wanted to accept from main I changed the script and pushed these changes as new commits. However, for the changes I didn't want to accept I removed the undesired option in the web editor, then committed these changes directly on GitHub. I then checked the status of my local branch, which was up to date (as I picked the option from my branch in the web editor).
> 
> My question now is what I would have to do in case I selected the other (main) option in the web editor. How would I then get my local branch up to date? Would that be by pulling from origin, and would that import the additional changes from the commit that was done directly on GitHub?

Yup - that's correct. You can resolve conflicts in either location. I generally pull everything down locally, merge main into my branch, resolve the conflicts and then push - when there are multiple files and things to compare, I find it easier to manage in VS Code. I do use the web editor for really simple cases. But as you've shown, you can also use a mixed strategy 😄

PR #42, opened 2025-04-09
Uploading era5 climate python script

lelavathy:
This Python script retrieves ERA5-Land climate data from the Copernicus Climate Data Store (CDS). It automates the download of high-resolution precipitation data, which is useful for hydrological modeling, climate studies, and ecosystem analysis.
 The script utilizes the `cdsapi` package to request ERA5-Land hourly total precipitation data for a specified year, covering all months and days.

'package_dependencies:
#'    - cdsapi
#'    - netCDF4
#'   - pandas
#'    - xarray

#' usage_notes: |
#'  - **Copernicus Data Store (CDS) Registration & API Key Setup**
#'    To access ERA5-Land data, users must register for a free account at the [Copernicus Climate Data Store (CDS)](https://cds.climate.copernicus.eu/).
#'    After registration, an API key is provided, which must be configured in a local `.cdsapirc` file for authentication.
#'    The file should be placed in the user’s home directory (`~/.cdsapirc` on Linux/macOS or `C:\Users\YourUsername\.cdsapirc` on Windows) and contain the following format:

#'    ```ini
#'    url: https://cds.climate.copernicus.eu/api/v2
#'   key: your-uid:your-api-key
#'    verify: 1
#'    ```

#'  - **Installing and Configuring the `cdsapi` Package**
#'    The `cdsapi` Python package is required to communicate with the Copernicus Data Store. Install it using:
#'    ```bash
#'    pip install cdsapi
#'    ```
#'    Ensure the `cdsapi.Client()` function is correctly initialized in the script before making any requests.

#'  - **Download Time Considerations**
#'    The dataset request spans an entire year (12 months, 365 days, 24-hourly data points per day).
#'    Due to the large data volume, download times vary based on:
#'      - Internet speed
#'      - CDS server response time (which may slow down during high demand)
#'      - The selected spatial extent (smaller regions will download faster)
#' ---

PR #87, opened 2025-07-23
Framework for grid definition and reprojection

davidorme:
This is a draft PR aimed at providing some structure for spatial grid and organising some of the spatial data. It

* Creates an `analyses/core` directory that contains the `maliau_grid_definition.py` and `maliau_grid_definition.toml`. The python file is **only** intended to document the decisions for creating the definition and the idea is that any code can read the TOML file to access information.
* Create an `analyses/abiotic/data_download` directory that contains three python files:
    * `cdsapi_downloader.py` and `maliau_era5_download.py` are a restructure of @lelavathy's files in #42 . The first file contains an importable function for grabbing the required ERA5 variables from CDS and then the second file just runs that function for 3x3 chunk around Maliau for 2010-2020. This data is in WGS84 projection.
    * `maliau_era5_reproject.py`: this is new functionality and shows how to use the `rioxarray` package to take the WGS84 data and reproject it onto a neat 90m grid in UTM50N.
    
None of these files are final - missing metadata for one thing - but gives us a starting point to discuss how to coordinate things like the spatial grid across module data collection efforts.


analysis/abiotic/data_download/cdsapi_downloader.py
```diff
@@ -0,0 +1,119 @@
+# noqa: D100
+
+#' ---
+#' title: Data downloader tools for ERA5
+#'
+#' description: |
+#'   This file defines a Python function that can be used to automates the download of
```
vgro, line 7:
```suggestion
#'   This file defines a Python function that can be used to automate the download of
```

analysis/core/maliau_grid_definition.py
```diff
@@ -0,0 +1,80 @@
+"""Maliau grid definition.
+
+This script is used to generate
```
vgro, line 3: This script is used to generate ??

vgro (commented):
This looks overall very good to me. A few minor points:

- I don't think `analysis/` is the right location for this, it might be better placed in a pre-processing directory
- we still don't seem to have a central location where the info about the grid is saved. The toml file holds the VE grid info but not the WGS84 coordinates, which is what you need to download ERA5, or did I miss something?

analysis/core/maliau_grid_definition.py
davidorme: Yeah - got lazy on the file metadata. Will fix this once any big picture stuff has shaken down.

analysis/core/maliau_grid_definition.toml
```diff
@@ -0,0 +1,122 @@
...
+]
+res = 90
+
+[core.grid]
+cell_area = 8100
+cell_nx = 50
+cell_ny = 50
+grid_type = "square"
+xoff = 494300
+yoff = 521300
```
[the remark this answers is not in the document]
davidorme: That's a good point - the coords are for the edges, not the cell centres, so those values should be `+45`

analysis/core/maliau_grid_definition.py
davidorme: Done

analysis/core/maliau_grid_definition.toml
```diff
@@ -0,0 +1,122 @@
+epsg_code = 32650
+min_x = 494300
+min_y = 521300
+max_x = 498800
+max_y = 525800
+cell_nx = 50
+cell_ny = 50
+cell_x_bounds = [
```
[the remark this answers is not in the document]
davidorme: I'm leaving them in for now but have converted to the more useful cell centres

analysis/core/maliau_grid_definition.toml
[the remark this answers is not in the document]
davidorme: Done

hrlai (commented):
Sounds good to me overall. Let's discuss the folder structure some day as mentioned in #66 

So, just to double check the general idea: 
1. a script outputs a TOML metadata for a site what we are working on together
2. each of us reads the TOML to output the input data for each module in the same spatial configuration (e.g., `maliau_era5_download_and_reproject.py`)

PR #89, opened 2025-07-24
object_name_linter

### LINKED ISSUE #71: `lintr` object_name_linter is not compatible with model coding

hrlai:
This is a reopen of #72 that was merged by mistake, and aims to resolve #71 .

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -2,5 +2,7 @@ linters: linters_with_defaults(
     cyclocomp_linter(complexity_limit = 10L),
     indentation_linter(indent = 2L),
     line_length_linter(length = 88L),
-    object_name_linter = NULL
+    object_name_linter(
+      regexes= "[[:alpha:]][[:alnum:]_]?[[:alnum:]]"
+    )
   )
--- a/analysis/soil/carbon_use_efficiency/update_cue_parameters.Rmd
+++ b/analysis/soil/carbon_use_efficiency/update_cue_parameters.Rmd
@@ -76,7 +76,6 @@ knitr::opts_chunk$set(
 )
 ```
 
-
 ```{r Libraries}
 # load packages
 library(tidyverse)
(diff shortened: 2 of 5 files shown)


hrlai:
@davidorme here's reopening the lintr PR. Sorry, I think I must have accidentally clicked "close with comment" rather than just "comment" last time!

I have copied your object name linter suggestion in #71 just to make a commit and create this PR.

davidorme:
@hrlai I've pushed a few changes to fix R names that failed the new specification. 

I also got carried away and did some formatting on one of your R markdown docs in order to wrap lines better and make text more readable. In the process, I discovered a new option for `knitr` that allows the chunk options to be set across multiple lines as YAML text.  If those changes seem OK, then I think we can merge.

davidorme (approved):

hrlai:
Thanks and overall looks good to me. However, the following does not seem to pass:
```
t <- litter$t
L <- litter$lignin
r <- exp(log_r)
```

I'd like to keep them because they match the single letter variable names in the mathematical equations... what do you think?

davidorme:
> Thanks and overall looks good to me. However, the following does not seem to pass:
> 
> ```
> t <- litter$t
> L <- litter$lignin
> r <- exp(log_r)
> ```
> 
> I'd like to keep them because they match the single letter variable names in the mathematical equations... what do you think?

Honestly, I think that is better avoided in code. In an equation, you'd have the text next to it that says "where $L$ is the lignin content etc." and the point of the short symbol is to make the equation compact. There's a _little_ bit of that in code, but in general it's easier to follow the code with more descriptive names.

hrlai:
Righto, I agree that having single-character object name is the exception rather than the norm. And if I absolutely must use a single letter, then there is always the #nolint option :-P

So let's use a lintr that enforce no single-letter object name then, i.e., we would avoid it unless absolutely necessary. Will merge and close this now, thanks!

davidorme:
I'd kinda argue that we _never_ need to use a single character variable name and there are good reasons not to. The `#nolint` option is there for when the linting rules suggests something actively bad  😄

PR #96, opened 2025-08-02
56 gitignoring data and globus

### LINKED ISSUE #20: Separate .gitignore in each analysis folder?
Sometimes an analysis script will generate large output files, should we gitignore them in each analysis folder, so we don't universally ignore them using the root .gitignore?

### LINKED ISSUE #56: gitignoring data and Globus
Thanks for the Globus sharing session. I could run it alright, but after syncing my local is now full of unstaged files, which I can ignore but they do clog up the git interface a lot... Eventually this list of unstaged files will get unmanageable?

I have two questions:
1. Are we going to add `data` or `data/primary` to gitignore sometime down the line? 
2. I uploaded primary data once, should I delete it from GitHub (by deleting it locally and then make a commit for it)?

hrlai:
This PR removes a data file that was pushed before we use Globus.


davidorme (approved):
👍

Issue #101, opened 2025-08-13
Estimate equilbrium litter stocks


jacobcook1995:
For some litter pools (e.g. wood) I believe that there are data on stocks from SAFE. But in other cases we only have data on litter fall rates. We can estimate litter stock equilibrium based on these litter fall rates and the decay rates previously found (see #60).

The equation for the rate of change of a litter pool is 

$\frac{dP}{dt} = I - D*P$,

where $I$ is the input rate to the pool and $D$ is the decay rate from the pool. As we are (for now) ignoring environmental effects, $D=k_m$ in the metabolic case and $D=k_s * e^{rL}$ in the structural case.

The value of $I$ is a bit tricker to calculate as it is an input to a specific pool (i.e. above metabolic) rather than just an input rate of leaves or fruit. However, if we assume that the leaf input has the same chemistry as the leaf litter used to calculate the decay rates we can just use the values for $f_m$ and $f_s$ that we previously calculated, i.e. the input to the above ground metabolic pool from leaves would be

$I_m = f_m I_l + I_r$

where $I_l$ is the leaf fall rate (in carbon mass terms), and $I_r$ is the fruit/flower fall rate (if we are assuming that fruit are all metabolic)

@hrlai , let me know if that all seems sensible

hrlai:
Thanks @jacobcook1995 they make sense.

Just checking the dimensions:
- $P$ in [mass per area]
- $I$ in [mass per area per time]
- $D$ in [per time], according to equations in #60 (?)
So I guess the units check out?

hrlai:
@jacobcook1995 should twigs / small branches be part of deadwood or aboveground structural + metabolic?

Issue #60, opened 2025-05-31
Model litter chemistry and turnover


hrlai:
See also https://virtual-ecosystem.readthedocs.io/en/662-improve-soil-nutrient-documentation/virtual_ecosystem/theory/soil/litter_theory.html#litter-chemistry

arne-exe:
> Currently I am using the SAFE [litter data](https://zenodo.org/records/3929632), which seems to be _leaf_ litter. In Elias et al. (2020) their methodology stated that coarse woody debris and partially decomposed leaves were removed. These leaf litter have N, P and lignin measured, so we can eventually link them to the metabolic and structural litter pools. But we do not have chemical data for root and reproductive litter.

Hey, I noticed you mentioned chemical data for root and reproductive litter, I'm currently working on stoichiometry in #48.
The following references may be of interest to you:
- Kitayama et al. (2015; DOI http://dx.doi.org/10.1111/1365-2745.12379), also see supplementary
- Aoyagi et al. (2018; DOI https://doi.org/10.1007/s11284-018-1642-9)
- Proctor et al. (1989; DOI https://doi.org/10.2307/2260752)
- Imai et al. (2010; DOI http://dx.doi.org/10.1017/S0266467410000350)

(Accidentally edited your comment instead of quoting it, my bad)

hrlai:
THANK YOU! I'll take a look 😄

hrlai:
Read and see if steady states in this paper makes sense https://besjournals.onlinelibrary.wiley.com/doi/full/10.1111/2041-210X.12138

jacobcook1995:
So for the decay of litter items that are assumed to have no lignin (i.e. fruit) the decay process should be quite straightforward, i.e. 

$$\frac{x(t)}{x_0} = e^{-k_m t},$$

where $k_m$ is the decay rate for metabolic litter. It gets a lot more complex for items that contain lignin as the model does two things:

1) Splits the litter into structural and metabolic based on lignin content
2) Calculates a factor that slows structural litter decay based on lignin content

The fraction of litter input that is considered to be metabolic is given by

$$f_m = f_M - L * (s_N * C:N + s_P * C:P),$$

where $f_M$ is the maximum fraction that can go into the metabolic pool, $L$ is the lignin content of the items, $C:N$ and $C:P$ are the stoichiometric ratios, and $s_N$ and $s_P$ are the "responsiveness" of the split to changes in stoichiometric ratio.

[...]

PR #104, opened 2025-08-27
102 rethinking litter decay models

### LINKED ISSUE #102: Rethinking litter decay models
While addressing #101 I estimated equilibrium root structural stock that is incredibly high. It was probably due to very slow decaying rate of this litter component. 

I then revisited the litter decay model in the merged PR #70 . On a closer look, I have a gut feeling that some of the parameters are "off"; it's not that they are wrong, it's just that they might have been strongly influenced by my prior choice. One of them is they key parameters $r$, which determines the effect of lignin on decay rates. It's really close to my prior, and I suspect that it has a lot to do with weak data going into the model --- lignin data is really crappy for the leaf and wood decomposition SAFE datasets I used.

In short, I think lignin is the bottleneck for us right now. 

I then tried to find datasets from scratch again, this time manually and using LLMs. No good yield. There simply wasn't enough studies that collected mass loss over time _and_ lignin per sample. 

Next step, I will scrutinise our model in #60 again, and think of ways to save parameters. My gut feeling is that some of them should be reparameterised together and then recovered post-fiiting with (lots of) assumptions, unfortunately.

hrlai:
This PR updates a previous analysis merged via #70 to better improve parameter estimation. When I say "improve" it is hard to say what has improved, because we really don't know what the expected values are. BUT, I think we can say that the improvement is in making sure that the fitted models are not as biased / sensitive to our prior choice (I think previously they were, as described in #102 .

What I did:
- Be less greedy from fitting everything in a single model, to fit three models: leaf litter decay, woody litter decay, and a lignin model. 
- Finding datasets with *both* lignin and decay data is tough, so I split them into separate tasks: first reparameterise (or absorb) lignin $L$ into the equations in #60 , and then recover them *a posteriori* when the third lignin model. This freed me up to find higher-quality datasets separately.
- The leaf litter model with two mixture components (i.e., $f_m$ and $1 - f_m$) was also another culprit causing inference issue, but I discovered that simply reparameterising the mixtures' summation with "log sums of exponentials" addressed most of the numerical challenges. 
- The leaf litter model allowed us to infer $k_m$, $K(L)$, (logit of) $f_M$ (more on this later), $s'_N = s_N \times L$ and $s'_P = s_P \times L$
- The woody litter model allowed us to infer $K_W$
- The lignin model allowed us to infer $r$

Then using average leaf and wood lignin from @arne-exe 's plant stoichiometry script, I recovered $`s_N = s'_N / L_\text{leaf}`$ , $`s_P = s'_P / L_\text{leaf}`$, $k_s = K(L) / \exp(-r L_\text{leaf})$, and $k_w = K_W / \exp(-r L_\text{wood})$. Assigning average leaf and wood lignin to everything will be a crude assumption that we need to accept at this stage.

I am setting this PR as a draft because I want to test it out on #101 and check if we still get crazy equilibrium values.


analysis/litter/chemistry_and_turnover/model_rate_lignin.R
```diff
@@ -0,0 +1,117 @@
...
+#| input_files:
+#|   - name: ecy201091123686-fig-0002-m.csv
+#|     path: data/derived/litter/chemistry_and_turnover
+#|     description: |
+#|       This csv file is digitised from Figure 1 in the original study, Geffen
+#|       et al. (2010) https://doi.org/10.1890/09-2224.1 using the metaDigitise
+#|       package in R.
+#|
+#| output_files:
+#|
```
annarallings, line 31: See comment above

analysis/litter/chemistry_and_turnover/model_leaf.R
```diff
@@ -0,0 +1,222 @@
...
+#|
+#| input_files:
+#|   - name: species+name+plant+traits+and+mass+loss+data+20190618.xlsx
+#|     path: data/primary/litter/
+#|     description: |
+#|       Litter chemistry and decomposition in subtropical broadleaf forest
+#|       by Tao et al. Downloaded from https://doi.org/10.5061/dryad.7hg8mp7
+#|
+#| output_files:
+#|
```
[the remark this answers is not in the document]
davidorme: In the usage notes or comments. The `output_files` metadata structure links to explicit files.

data/derived/litter/chemistry_and_turnover/ecy201091123686-fig-0002-m.csv
annarallings, line 1: For consistency, I'd avoid loading this data in GitHub (though it is really small)

annarallings (commented):
Just a couple comments on documentation and data - looks good otherwise.

analysis/litter/chemistry_and_turnover/model_leaf.R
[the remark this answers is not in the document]
hrlai: @annarallings and @davidorme there is no output file from this script, so I left it blank. Do you mean saying this explicitly in the usage notes? 

And if there is no output file, do you prefer us to left the section blank, or remove it entirely?

data/derived/litter/chemistry_and_turnover/ecy201091123686-fig-0002-m.csv
hrlai: @annarallings this dataset is extracted from a figure (which is considered "primary"), so I save the extracted data as "derived". We haven't decided yet, but the current feeling is to avoid pushing primary data at all (meaning the folder will probably be removed entirely). However, I am not 100% sure about derived data like this. Do you want me to treat it more like a primary data and not push it (use Globus) instead?

analysis/litter/chemistry_and_turnover/model_leaf.R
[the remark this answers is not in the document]
hrlai: Ok I'll leave a note for now.

analysis/litter/chemistry_and_turnover/model_rate_lignin.R
hrlai: Resolved as the above.

arne-exe (approved):

PR #109, opened 2025-09-18
Free-living nitrogen fixation model

### LINKED ISSUE #83: Parameterise free-living N fixation
The equation for free-living nitrogen fixation rate is 
$$ F_r Q_{10}^{(T_s - T_r)/10} \,, $$
where $F_r$ is the fixation rate at reference temperature, $Q_{10}$ the Q10 temperature quotation, $T_s$ the soil temperature, and $T_r$ the reference temperature. This is Equation set 10 from Lin et al. (2000).

Lin et al. took values from Granhall et al. (1981), which are assumed and not empirically parameterised. I will try to do the latter.

hrlai:
This is a draft PR. The goal is to parameterise the free-living N fixation model, but let's do some sanity check before proceeding. 

@jacobcook1995 I fitted the top-10 models in Kontopoulos et al. (2024), and the Eubank model turns out to fit the dataset best. The [dataset ](https://doi.org/10.5066/P1MFBVHK)I obtained is _very_ recently published, then I filtered to use only free-living N fixation. I then paired it with a [global map of soil temperature](https://doi.org/10.5281/zenodo.7134169).


hrlai:
What I wanted to check with you is the following prediction graph:

image

Blue is the fitted curve, and dots are the data (they are not well sampled at higher soil temperatures; very few tropical datasets). 

The Eubank equation does infer a peak temperature, which was estimated as 21.1C. It is a bit on the lower end of what the literature says?

jacobcook1995:
@hrlai , that all seems sensible to me!

I guess the bottom line is that we have limited wriggle room in terms of increasing model complexity and I am reluctant to spend it on this specific issue. So, I think the decisions you have made to keep the function complexity down are very much justified, even if they might be a bit dubious if modelling nitrogen fixation was our singular focus

jacobcook1995 (approved):
This looks good!

One thing we probably should discuss at some point is how/when we start integrating these improved functions and parameter estimates in the `virtual_ecosystem`. But I guess that's something to talk in a meeting rather than via Github

hrlai:
> how/when we start integrating these improved functions and parameter estimates in the virtual_ecosystem

Yeah it feels that we have reached a critical mass to start updating VE values. Let's discuss during the next meeting :)

annarallings (approved):
Sorry I didn't realize I missed this.

PR #114, opened 2025-09-22
Initial value for soil ammonium and nitrate

### LINKED ISSUE #113: Model soil ammonium and nitrate
Continuing from [#62 ](https://github.com/ImperialCollegeLondon/ve_data_science/issues/62#issuecomment-3315275309)

hrlai:
This is draft PR to estimate the initial values of soil ammonium and nitrate. 

@jacobcook1995 two things to check with you:
1. The [dataset ](https://zenodo.org/records/3258117)I used had negative values sometimes, can I assume that they are zero concentrations?
2. I estimated an average ammonium of 3.9 kg/m3 and nitrate of 5.0 kg/m3. They are magnitudes larger than the fake initials in VE's example data. What do you think?


jacobcook1995 (approved):
LGTM!

Yeah in terms of the negative values, I can't see anything in the paper they link that mentions the negative values. I guess they must be measurement errors which they preserved so that the underlying variability isn't underestimated (assuming the machine is equally likely to misestimate in either direction). I guess we probably want to treat negatives as zero if we are trying to get a spatial map as we can't allow zero values. But I guess if we wanted a landscape average or similar we would want to keep the negative values in

annarallings (approved):
LGTM

PR #126, opened 2025-10-07
Update README.md

nickwctan:
Fixing URL to github overview and getting started.


davidorme (approved):
This is great - thanks for the update. I've added the last link as well.

PR #155, opened 2025-11-25
132 Add remaining plant input data

### LINKED ISSUE #132: Provide update summary on plant input data
This issue focuses on preparing an update summary on plant input data.
-verify and update plant model data needs
-evaluate which parameters are representative for Maliau so far
-if not representative for Maliau, evaluate broader representativeness of parameters

arne-exe:
This PR adds the remaining plant input data:

Still to add:
~~-Reproductive tissue specific respiration rate (add to PFT csv)~~
~~-Distinguish senesced leaf stoichiometry and lignin from the other foliage (add to PFT csv; note that parameter names have changed from senesced_leaf to leaf_turnover for stoichiometry; senesced leaf lignin is still in plant constants)~~
~~-per_stem_annual_mortality_probability (add to plant constants)~~
~~-per_propagule_annual_recruitment_probability (add to plant constants)~~
~~-root exudates (add to plant constants)~~
~~-plant_pft_propagules (I have this in plant model data needs but can't seem to immediately spot where/how it is being used in the model)~~

-subcanopy seedbank stoichiometry and lignin (currently uses the same values as for tree reproductive tissues)

This PR also adds a new folder in the derived plant data folder, called csv_plant_input_data (as well as a corresponding analysis folder containing the scripts that generate the files in csv_plant_input_data ).
The purpose of the csv_plant_input_data folder is similar to the netcdf_plant_input_data folder, in that it contains the "final" input data files to be used in specific model runs. The advantage of having a script that prepares a final/clean version of these input files is so that entire scripts do not need to be run again if for example the name of a variable changes in the model. Also, some of the constants are provided in the same output file as for the T model parameters. While the values for these constants need to be available in order to prepare the configuration, they should not be included in the final plant_functional_types CSV file, so that's another reason for using the csv_plant_input_data folder and related "analysis" scripts.

As mentioned in (https://virtual-ecosystem.readthedocs.io/en/1104-document-new-configuration-system/using_the_ve/science_model_configuration.html#plants-model), the CSV files within this folder should be:
- a CSV file defining the plant functional types to be used in the model (which will be accessed in the configuration through [plants.pft_definitions_path])
- a CSV file defining the size-structured communities of plant functional types found in each cell (accessed in the configuration through [plants.cohort_data_path])
- potentially a CSV file for the constants, so that they are organized neatly together. If I understood correctly, this CSV file will not be accessed through the configuration, instead the user needs to manually specify the constants in the config.


arne-exe:
@davidorme root exudates in the plant constants were set to 0.5 (=50%) of GPP. I've now calculated these at around 0.02 (2%).
This could have been another possible reason why trees were not growing before (as basically half their available carbon was removed?).

davidorme:
> @davidorme root exudates in the plant constants were set to 0.5 (=50%) of GPP. I've now calculated these at around 0.02 (2%).
> This could have been another possible reason why trees were not growing before (as basically half their available carbon was removed?).

@jacobcook1995 I have a memory that this was set from some soil related research?

arne-exe:
-leaf_lignin was calculated for each PFT but only one value can be assigned in the constants. Cases like this highlight where this specific constant would better fit as part of the pft_definitions. 
-Same for senesced_leaf_lignin.
-I've picked the value for emergent trees for now

annarallings (approved):
LGTM

PR #162, opened 2026-01-14
# 162 Adjust cell_id and time dimensions for Maliau plant input data

arne-exe:
Added cell_id dimension as variable so that it correctly counts from 0:249 instead of 1:250.


arne-exe:
Updated NetCDF file has also been added to Globus.

arne-exe:
> As part of this PR (and to match #164), I will also check that the cohort distribution uses 1 ha cells (100m x 100m).

plant_functional_type_cohort_distribution_Maliau_50x50 has been updated as the original version was rescaled to 8100m2 (so the scaling is undone and the basic per hectare version is used).

Note that subcanopy_vegetation_biomass and subcanopy_seedbank_biomass do not need to be scaled, as their unit is kg C m-2 (so they are not directly affected by changes in cell area).

PR #168, opened 2026-01-26
#167 Adding simplified PFT input data Maliau

### LINKED ISSUE #167: Prepare simplified PFT input data for comparison
As discussed with @annarallings, I should prepare a simplified PFT version for the Maliau input data to allow for comparison with the "base" version.
Simplified PFT input data will only contain overstory and understory trees (so no emergent or pioneer trees).
Understory vegetation will remain the same.

The PFT traits data can remain as is but the input data file should be a subset only containing the 2 PFTs of interest.

The cohort distribution will need to be changed.
There are 2 options here, either use the base cohort distribution and remove all trees belonging to PFTs that are not of interest, OR rerun the cohort distribution preparation script and replace the trees that are not of interest with the PFT that are of interest.
I think the latter option makes more sense for the intended purpose of this simplified PFT version (i.e., we want to know how complexity changes things in the model). Removing the PFTs not of interest would change the complexity but would also remove half the biomass, so I think replacing just the PFT but keeping the individuals will more closely track what's going on in the system.

The NetCDF input data will also need to be changed, as the PFT dimension will not longer match.

arne-exe:
This PR prepares/adds simplified PFT input data for Maliau.

### DIFF
--- a/analysis/plant/csv_plant_input_data/csv_plant_input_data_Maliau_50x50_simplified.R
+++ b/analysis/plant/csv_plant_input_data/csv_plant_input_data_Maliau_50x50_simplified.R
@@ -56,7 +56,7 @@
 #|       This CSV file contains the plant constants for Maliau.
 #|
 #| package_dependencies:
-#|     - XX
+#|     - tidyverse
 #|
 #| usage_notes: |
 #|   This script prepares the final version of the plant input data for Maliau.
--- a/analysis/plant/csv_plant_input_data/csv_plant_input_data_Maliau_50x50_simplified.R
+++ b/analysis/plant/csv_plant_input_data/csv_plant_input_data_Maliau_50x50_simplified.R
@@ -102,6 +102,18 @@ subcanopy_parameters <- read.csv(
 
 plant_pft_definitions_Maliau_50x50 <- t_model_parameters # nolint
 
+# Exclude root_exudates, per_stem_annual_mortality_probability and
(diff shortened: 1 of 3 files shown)


analysis/plant/csv_plant_input_data/csv_plant_input_data_Maliau_50x50_simplified.R
```diff
@@ -0,0 +1,346 @@
...
+#|     path: data/derived/plant/csv_plant_input_data
+#|     description: |
+#|       This CSV file contains the plant cohort distribution for Maliau.
+#|   - name: plant_constants_Maliau_50x50_simplified.csv
+#|     path: data/derived/plant/csv_plant_input_data
+#|     description: |
+#|       This CSV file contains the plant constants for Maliau.
+#|
+#| package_dependencies:
+#|     - XX
```
annarallings: tidyverse

annarallings (changes requested):
Just one small change but otherwise LGTM

arne-exe:
I realised that 50 by 50 grid is actually 2500 cells, not 250.. 
So will need to change this first before merging.

arne-exe:
Script updated and files added to Globus.

arne-exe:
Script updated and files added to Globus.

arne-exe:
Script updated and files added to Globus.

annarallings (approved):
LGTM

PR #176, opened 2026-02-06
174 Removed 3 constants from pft_definitions

### LINKED ISSUE #174: Remove root exudates, mortality and recruitment from pft_definitions
In https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/1314 I noticed that ```root_exudates ```, ```per_propagule_annual_recruitment_probability``` and ```per_stem_annual_mortality_probability``` are currently provided through the pft_definitions CSV file as input data. 
This is not correct as they are constants.
Loading them from within the pft_definitions file causes an error because the variables are not expected here.
So, need to update the input data accordingly.

arne-exe:
Fixes #174 by removing 3 constants from pft_definitions.


annarallings (approved):
LGTM

PR #195, opened 2026-03-04
193 estimate soil C fraction in POM and MAOM

### LINKED ISSUE #193: Estimate soil C fraction in POM and MAOM
Emma Sayer shared their data. First thing is to quantify the C fraction in POM and MAOM. They only looked at these two pools using fractionalisation. That's ok as a first step as these pools are the dominant ones. 

Second step is to look at the remaining proportion, which can be assigned to LMWC, microbes and necromass. This will be much trickier.

hrlai:
This PR fits two models, one to estimate POM and MAOM carbon content from total soil carbon, while another model does the equivalent for nitrogen. 

The outputs are two model objects to be used later for post-hoc predictions.


jacobcook1995 (approved):
Looks like a sensible approach to me given the current lack of data.

(probably worth chatting about the current situation with SAFE carbon fractionalisation data in our meeting this week)

PR #205, opened 2026-03-05
Update elevation script to use cell_id structure

lelavathy:
This update modifies the Maliau elevation preprocessing script so that the
output dataset follows the Virtual Ecosystem (VE) spatial data structure and
addresses the issue reported in “Input data is flipped when mapped on grid
#1054” #1054

Key changes include:

Replacing the 2D spatial dimensions (x, y) with a single cell_id
dimension consistent with VE grid structure.

Flattening the elevation grid using row-major ordering with a top-left
origin (cell_id = 0 at the top-left cell).

Explicitly controlling the orientation of the y-axis to prevent the grid
from being flipped when mapped to VE cells.

Storing per-cell spatial coordinates (x, y, and UTM coordinates) so
that each cell_id can be correctly mapped back to its spatial position.

These changes ensure that the elevation dataset aligns with the VE grid
layout and prevent the input data from being vertically flipped when mapped
onto the simulation grid.

### DIFF
--- a/analysis/abiotic/data_download/maliau_elevation_data_processing_script.py
+++ b/analysis/abiotic/data_download/maliau_elevation_data_processing_script.py
@@ -183,100 +183,65 @@ if np.isnan(dst_data).any():
     dst_data = filled_data
 
 # Reformat DEM to Virtual Ecosystem (VE) 1D grid structure
-# This section converts the processed elevation dataset from
-# a 2D spatial grid (x, y) into the VE cell-based format.
-# The elevation grid is flattened so that each spatial grid cell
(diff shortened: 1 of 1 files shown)


vgro:
@lelavathy it looks like this is no longer relevant and a new issue has been created, if you agree could you close the pull request please?

lelavathy:
> @lelavathy it looks like this is no longer relevant and a new issue has been created, if you agree could you close the pull request please?

Thanks, I agree this is no longer relevant. I’ll close this PR and open a new one to update the x and y coordinates using the correct UTM Zone 50N values.

PR #210, opened 2026-03-05
201 bacteria and fungi ratio in soil carbon

### LINKED ISSUE #201: Bacteria and fungi ratio in soil carbon
We need to split soil microbial C pool into bacteria and fungi. Previously we modelled fungi:bacteria ratio in #57 based on PLFA, which is a proxy for biomass. So the next step is simply to convert biomass ratios into C ratios --- I will use stoichiometry database for this.

hrlai:
This PR uses the main script `analysis/soil/nutrient_pools/carbon_microbial_guild.R` to take outputs from
1. `analysis/soil/fungi_bacteria_ratio/model.R` (old, with minor updates)
2. `analysis/soil/fungi_guild_ratio/model_SAFE.R` (old, with minor updates)
3. `analysis/soil/stoichiometry/bacteria_fungi.R` (new script)

and then combine them to calculate how soil microbial carbon (extracted from `analysis/soil/nutrient_pools/carbon_microbial.R`) should be split into bacteria, saprotrophic, AM and EM fungi.


analysis/soil/stoichiometry/bacteria_fungi.R
```diff
@@ -0,0 +1,92 @@
...
+  pmap(name_backbone, .progress = TRUE) |>
+  bind_rows()
+C_frac <-
+  query |>
+  select(kingdom) |>
+  bind_cols(stoich) |>
+  filter(!is.na(kingdom)) |>
+  select(kingdom, C_mean) |>
+  # convert C fraction from percentage to proportion
+  mutate(C_mean = C_mean / 100)
```
jacobcook1995: I'm guessing split out fungi by type wouldn't be possible as mycorrhizal types are not monophyletic? (don't think that's particularly a problem if so, mycorrhizal fungi do have different stochiometries but I imagine that's mainly to do with the amount of nutrients they have, which are a small component of biomass, rather than how much of their mass is carbon vs water)

jacobcook1995 (approved):
All looks sensible to me from a science perspective (can't really comment on code style as `R` is really not my language)

analysis/soil/stoichiometry/bacteria_fungi.R
hrlai: It is possible and I tried to do it in a new commit. There was an imbalance in the number of sampled EM vs saprotroph in this stoichiometric database (148 EM vs. 77 saprotroph, and no AM unfortunately). The estimated carbon content of EM was lower than that of saprotroph but only slightly so:
```
        EM saprotroph   bacteria 
 0.4366595  0.4375179  0.5000000
```

I'll merge this PR for now and treat this as something good to have for the future?

PR #211, opened 2026-03-06
Estimate fungal fruiting biomass

### LINKED ISSUE #121: Fungal fruit biomass

hrlai:
This PR estimate fungal fruiting biomass to initialise the Maliau scenario, in a very crude way! See description in the R script's metadata for more details. There is only one file in this PR.

I decided to stop here because I have spent too much time hunting down this variable and it's becoming hard to justify more time, at least for the purpose of Maliau initialisation.

@jacobcook1995 the value here is one order of magnitude lower than the soil example data value, but I think it's okay since there is no right or wrong answer now...


jacobcook1995 (approved):
Yep a crude estimate is completely justified here (it's actually really interesting that it's such a big data gap).

Wouldn't worry about comparing to the previous value as that was a total guess (I couldn't even find existing models with fungal fruit)

PR #213, opened 2026-03-06
77 input real soil data from safe

### LINKED ISSUE #77: Input real data from SAFE
In the next step let's try to input real soil data from SAFE (?) This will be a hard test like what @lelavathy is doing with the hydrology.

For many soil variables we probably won't have spatially explicit data, but here's a first pass:
- When data are spatially implicit, model mean and variance and simulate them randomly across grids
- When data are spatially explicit, include a spatial autocorrelation term in the model and then in the simulation

### LINKED ISSUE #122: Joint modelling of soil variables with incomplete covariance
This is currently a note-to-self on soil data integration from multiple datasets.

We will be facing a common issue as follows:
1. In Dataset A we can fit a multivariate model to soil variables $X_1$, $X_2$, ..., $X_n$ and estimate their covariance
2. In Dataset B we can fit another multivariate model to soil variables $X_1$, $Z_1$, $Z_2$..., $Z_m$ and estimate their covariance
3. From above, we get an incomplete covariance matrix for $\mathbf{X}$ and $\mathbf{Z}$, linked only through the shared variable $X_1$

Potential ways forward for making multivariate predictions:
1. Assume conditional independence between the unshared variables
2. Estimate the full covariate matrix _if the model can take NA values_

Option 2 is much more ideal, as I think we can do this through CBFM.

hrlai:
This PR sets up a statistical modelling pipeline to generate initial values for the Maliau scenario (#77). 

@jacobcook1995 this is related to our discussion on the spatial soil modelling.
@arne-exe and @nickwctan I just need one of you to review my R codes, and same for any future requests. Please feel free to assign me to your own tasks 😄 

The goal here is not to have a final model, but a WIP model that is merged into main so other branches can test it out (this WIP model is intended to be paired with lots of post-hoc downstream analyses to, for example, split total carbon into separate carbon pools).

There are two scripts:
1. `analysis/soil/initialisation/model_danum.R` This is a not intended for future use, just an exploration with the Danum side of things (since they have sent me the plot data)
2. `analysis/soil/initialisation/model_safe.R` This is the main script going forward. It is named SAFE because the main dataset is the SAFE soil campaign, though the prediction output is for Maliau.


analysis/soil/initialisation/model_safe.R
```diff
@@ -0,0 +1,284 @@
...
+    Y = maliau$cell_y_centres
+  )
+
+maliau_dat <-
+  maliau_grid %>%
+  mutate(
+    elev = extract(elev, .[, c("X", "Y")])[, "SRTM_UTM50N_processed"],
+    topo = extract(topo, .[, c("X", "Y")])[, "SRTM_UTM50N_TRI_Wilson2007"],
+    hydro = extract(hydro, .[, c("X", "Y")])[, "SRTM_Log_Flow_Accum"],
+    # acd = extract(acd, .[, c("X", "Y")])[, "Danum_acd"],  # nolint
```
jacobcook1995: This is commented out, but surely it should be Maliau ACD?

jacobcook1995 (approved):
Looks great on the science side, just had one minor query

analysis/soil/initialisation/model_safe.R
hrlai: ah yes, this should be Maliau, but the name is confusing due to the merging of Danum-Maliau-SAFE layers at the [beginning of this code](https://github.com/ImperialCollegeLondon/ve_data_science/pull/213/changes/BASE..96e036793c367edcf735ca56a1017d77f51ddfd5#diff-9fd4a9a9863bf8474cd83daa666a156727f393e86cca8d80696e874b9b6c538fR121-R126). Because the Danum layer came first, the merged output took the name of the Danum layer. I have renamed it for clarity.

arne-exe (approved):

PR #216, opened 2026-03-09
204 Upload tree demography validation script

### LINKED ISSUE #204: Tree demography
Are trees able to survive, and do tree demography outputs reflect mortality and recruitment correctly? 

Calculate tree mortality during the simulation from n_individuals per cohort in the outputs. Then compare this to expected tree mortality when applying the `annual_mortality_propability` to the cohort distribution. 

The next step is then to compare this mortality with real values. I could compare model outputs with different census years in the SAFE tree census, but census data seemed to vary quite a lot across years (i.e., not every year measured the same trees). So, calculate the proportion of trees that died in each census year, and compare these estimates with the one obtained from the model. 

To validate recruitment, I should compare the number of new recruits (calculated from n_individuals per cohort in the outputs) with estimates of new recruits from existing studies.

Note that the probabilities of mortality and recruitment are not PFT specific (they are plant constants). So although the amount of trees dying per PFT may be different due to differences in the initial cohort distribution, the relative mortality should be the same across PFTs. So keep that in mind when processing validation data.

arne-exe:
Validation script (work in progress) for demography (mortality and recruitment).

Calculated recruitment still needs to be compared to existing studies.
Rerun the script with updated simulation outputs, in particular a simulation with updated plant constants for mortality and recruitment.


arne-exe:

PR #218, opened 2026-03-09
Calculate soil P fractions for initialisation

### LINKED ISSUE #188: Soil initialisation: P pools

hrlai:
@jacobcook1995 I want to do a sanity check before making a full PR. 

In the single script here I have estimate the proportion of soil P in each pool to be (this is the final output from the single R script): 
```
  fraction                amount    prop
  <chr>                    <dbl>   <dbl>
1 soil_p_pool_dop           94.8 0.0288 
2 soil_p_pool_labile        35.5 0.0108 
3 soil_p_pool_particulate   10.6 0.00321
4 soil_p_pool_maom         201.  0.0609 
5 soil_p_pool_secondary   2968.  0.901  
6 soil_p_pool_primary       15.1 0.00458
```

Can you take a look at the `prop` column and let me know if they make sense? Do we expect secondary inorganic P to be largest pool, by far?


jacobcook1995 (approved):
Looks sensible to me! Not surprised that this is an area with limited data

PR #219, opened 2026-03-09
206 Added dbh and height growth validation script

### LINKED ISSUE #206: Tree dbh and height growth
Are trees growing at the right pace in terms of dbh and height?

I should compare VE diameter and height growth with existing studies. The input data for dbh and for the relationship between dbh and height were derived from the SAFE tree census dataset, however, only using a particular year where both dbh and height were measured. I could use the values from a different year from the same SAFE dataset as this would enable having PFT specific average dbh growth rates as validation data.

Regarding height, the input data for the relationship between dbh and height uses Maliau specific data but these height estimates were (likely) based on visual estimates, so may not be 100% accurate. It would be good to evaluate the predicted height gain over time by comparing it with other studies.

arne-exe:
Validation script for dbh and height growth (work in progress).


arne-exe:

PR #221, opened 2026-03-09
reorganise `data/site` folders

### LINKED ISSUE #220: Reorganise `data/site` folder
Following a round of discussion with the data team, @lelavathy and I are proposing the following minor folder restructuring for site-specific files:
- move files like `maliau_grid_definition_100m.toml` to `data/derived/site`; basically we don't keep `data/site` as a separate entity and classify it as either primary or derived data, like the VE modules
    - we also considered putting this file in `data/scenarios/maliau` but the above feels better
- store primary site files like the SAFE gazetteer in `data/primary/site`

These files will be committed and pushed to github, not via Globus anymore, so we can track changes.

hrlai:
This PR is a minor reorganisation of how we store site specific data and scripts, as discussed in #220 

@lelavathy could you check if this matches our decision?

@nickwctan @Baizurah @arne-exe tagging you all, review is not necessary. These files are mostly used by me and Lela only but making sure that you're kept in the loop. I guess in the future you might find some use of these folders.


lelavathy (approved):
@hrlai Yes, this matches our decision. The structure looks much clearer now 👍 Thanks for updating it.

PR #223, opened 2026-03-10
Estimate soil DON and DOC

### LINKED ISSUE #202: Estimate soil LMWC

### LINKED ISSUE #222: Estimate soil dissolved organic nitrogen (DON)

hrlai:
This PR estimates DON and DOC for soil initial data. See the single R script for assumptions.

Something worth repeating here is that we will use DOC as an imperfect proxy for LWMC, for the purpose of getting as close to a Maliau scenario as possible. 

@jacobcook1995 there is a also small question about bulk density in the script for discussion.

### DIFF
--- a/analysis/soil/nutrient_pools/doc_don.R
+++ b/analysis/soil/nutrient_pools/doc_don.R
@@ -34,9 +34,9 @@
 # values are in units [microgram / gram]
 
 doc_mean <- (117.6 + 102.2) / 2
-doc_sd   <- sqrt(11.2^2 + 7.8^2) / 2
+doc_sd <- sqrt(11.2^2 + 7.8^2) / 2
 don_mean <- (50.7 + 63.5) / 2
-don_sd   <- sqrt(14^2 + 8.8^2) / 2
+don_sd <- sqrt(14^2 + 8.8^2) / 2
 
 # Placeholder code to simulate DOC and DON across grids
 # then convert units to kg/m^3 assuming a bulk density of 1400.0 kg/m^3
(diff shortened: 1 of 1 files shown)


analysis/soil/nutrient_pools/doc_don.R
```diff
@@ -0,0 +1,52 @@
...
+
+doc_mean <- (117.6 + 102.2) / 2
+doc_sd <- sqrt(11.2^2 + 7.8^2) / 2
+don_mean <- (50.7 + 63.5) / 2
+don_sd <- sqrt(14^2 + 8.8^2) / 2
+
+# Placeholder code to simulate DOC and DON across grids
+# then convert units to kg/m^3 assuming a bulk density of 1400.0 kg/m^3
+# (this bulk density comes from the soil microbe config; or should we use
+# the current abiotic module constant 1175.0 kg/m^3 ?)
```
jacobcook1995: Definitely wouldn't use 1400.0 as that's very much a toy value I used to quickly estimate some parameters. The abiotic model constant has an [actual source](https://www.nature.com/articles/s41597-022-01481-5) for their value, so certainly should be preferred.

The soil campaign also estimated bulk density.

Guess it depends a bit on what you doing which you use as the abiotic model constant is the bulk density value that the model is using, whereas the soil sampling campaign bulk density is the relevant value for that dataset?

jacobcook1995 (changes requested):
Generally looks like a sensible approach, but agree that we need to make a decision about bulk density

analysis/soil/nutrient_pools/doc_don.R
hrlai: ah yes, of course we should use SAFE value. I have updated it. Currently it's hard coded in, but this is only temporary, because when I compile the soil initial data this part of the script will be moved to a master script.

jacobcook1995 (approved):

PR #230, opened 2026-03-12
enforce positive bound on nutrient values using half-Normals

### LINKED ISSUE #229: Litter model fails due to invalid lignin_below_structural values in litter_maliau.nc
Following the recent update, I tested the latest setup using the new litter input dataset litter_maliau.nc, which now uses x and y dimensions instead of cell_id. @hrlai 

The run fails with the following error:
``[ERROR] - litter_model - _setup(231) - Lignin proportions not between 0 and 1 found in: lignin_below_structural``

and followed by: 
``[CRITICAL] - main - initialise_models(84) - Configuration failed for models: litter``

This occurs during the litter model setup stage, so I checked the variable `lignin_below_structural` directly: 
min = -0.0471118725836277
max = 0.7474052309989929

Likely the error triggered by lignin proportions that is below 0. 

VE Version: 0.1.1a15

hrlai:
Thanks @Baizurah for picking this error out. I have updated the script, and the new litter input data is now transferred to Globus. Please test it out and approve this PR if it runs okay.

### DIFF
--- a/analysis/site/maliau_litter_data.R
+++ b/analysis/site/maliau_litter_data.R
@@ -182,11 +182,13 @@ source("analysis/litter/nutrient_pool/initial_nutrient_belowground.R")
 # simulate predictions
 below_litter_sim <-
   # first generate random C, N, P and lignin values
+  # using abs(rnorm(...)) to generate half-Normal random variates so that
+  # nutrient values are positive bound
   data.frame(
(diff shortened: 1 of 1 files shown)


Baizurah (approved):
I've tested the latest change to this input data, and it worked. Thanks.

PR #231, opened 2026-03-13
Guesstimate enzyme concentration

### LINKED ISSUE #200: Enzyme C pools
Most of the literature examined enzyme _activity_ rates not abundance.

hrlai:
This PR resort to a crude guesstimate of enzyme concentration for the Maliau initialisation because these variables are simply too hard to find empirically. @jacobcook1995 the reasoning and thought process are all in the single script of this PR. They don't do much and will be used in the master script that generates initial soil values.


jacobcook1995 (approved):
PR looks sensible, I agree that if you are only finding activities in the literature we need to use a simple assumption instead

PR #232, opened 2026-03-13
225 generate maliau soil netcdfs

### LINKED ISSUE #225: Generate Maliau soil netCDFs

hrlai:
Adds initial data for the soil module for the Maliau scenario. The main script is `analysis/site/maliau_soil_data.R`, and other scripts received minor changes (mostly to make them compatible with the main script).

@jacobcook1995 basically the same idea as the litter module, would be nice to get you to go over the assumptions, rationales or train of thoughts.


arne-exe (approved):

analysis/site/maliau_soil_data.R
```diff
@@ -0,0 +1,509 @@
...
+  ) |>
+  mutate(
+    elev = (elev - mean(soil$elev)) / sd(soil$elev),
+    topo = (topo - mean(soil$topo)) / sd(soil$topo),
+    hydro = (hydro - mean(soil$hydro)) / sd(soil$hydro),
+    acd = (acd - mean(soil$acd)) / sd(soil$acd),
+    evi = (evi - mean(soil$evi)) / sd(soil$evi)
+  ) |>
+  # we need to fill in two NA grids in the EVI layer,
+  # I think they are due to rivers / water bodies
```
jacobcook1995: Probably worth discussing this in our meeting on Friday, this could be due to the satellite not getting any cloud-free images for a specific pixel in the time period the mosaic is averaged over (3 months). It's possible we could fix this by using multiple quarterly mosaics, but that comes with issues that we should think through

analysis/soil/initialisation/model_safe.R
```diff
@@ -228,58 +226,11 @@ fitcbfm <-
...
 # some exploratory statistics
-coef(fitcbfm)
+# nolint start
 
-corrplot::corrplot(corB(fitcbfm), order = "FPC", diag = FALSE, type = "lower")
-corrplot::corrplot(corX(fitcbfm), order = "FPC", diag = FALSE, type = "lower")
+# coef(fitcbfm)
 
+# corrplot::corrplot(corB(fitcbfm), order = "FPC", diag = FALSE, type = "lower")
+# corrplot::corrplot(corX(fitcbfm), order = "FPC", diag = FALSE, type = "lower")
```
jacobcook1995: I would just delete these instead of commenting them out as they will still be stored in the git history anyway

analysis/soil/nutrient_pools/carbon_microbial_guild.R
```diff
@@ -57,6 +57,12 @@ fungi_rel_abun <-
 # Fungal and bacterial carbon fraction
 C_fraction_microbe <-
   read_rds("data/derived/soil/nutrient_pools/C_fraction_microbe.rds")
+# AM fungi don't have carbon value, so we assume it to have the average of
+# EM and saprotroph
```
jacobcook1995: Seems like the best assumption we can make (don't honestly know enough about fungal physiology to say whether AM fungi are more similar to EM fungi or saprotrophic fungi)

jacobcook1995 (approved):
The overall approach and assumptions seem reasonable to me. And as we agreed before, the best approach is to decide what needs further refinement during calibration

PR #236, opened 2026-03-17
207 Adding tree carbon cycling validation script

### LINKED ISSUE #207: Tree carbon cycling
Are trees cycling carbon realistically? This includes looking at:
1) Tissue standing carbon mass (i.e., in stem, leaf, root and reproductive tissues)
Do values make sense at tree, PFT and cell level? Also evaluate tissue masses relative to each other (e.g., leaf to stem).
Also evaluate changes in tissue standing carbon mass over time, to assess allocation to plant organs.

2) Tissue respiration
Need to check if using tissue specific respiration fractions produces realistic values.

3) Tissue litter production
This will heavily rely on tree productivity but need to check if using tissue turnover values produces realistic values.

4) Tissue animal consumption (particularly leaves)
This one is more related to animal model but I've included it as it's part of carbon cycling.
Also, the validation dataset for 4) is the same as 1) and 3) so might as well check it.

5) Plant symbiote carbon supply and root carbohydrate exudation
Need to check if values make sense, input data for these 2 variables may need calibration.

arne-exe:
Adds validation script (work in progress) for tree carbon cycling components.


arne-exe:

PR #240, opened 2026-03-19
103 additional files for gitignore

### LINKED ISSUE #103: Additional files for .gitignore
Hi data team, please feel free to add file extensions that we should add to `.gitignore` here. I think we can collect a bunch of them and then make a single PR, rather than many tiny ones. 

My directory is getting messy with uncommited files!

As a start, I have 
- `*.jpg` for digitised data.
- can also ignore all `.csv` in `data/primary` ???

hrlai:
This PR populates `.gitignore` to ignore:
- More R and RStudio auxillary files
- More VS Code auxillary files, except `.vscode/` which existed from the start
- Image data files
- `data/primary` but leaving a `.gitkeep` to maintain its existence (mainly to make it clear that there is this folder to sync via Globus)

Data team please feel free to add more files to ignore, otherwise feel free to review and merge.


arne-exe:
> > In #259 I noticed that some of the data files in the derived data folder are being tracked, which is not necessary as they are uploaded and updated through Globus.
> 
> Would you be able to list the files here?

They're all .csv files in data/derived/plant.
The ones I'm aware of that are being tracked are:

data/derived/plant/plant_functional_type
- plant_functional_type_cohort_distribution.csv
- plant_functional_type_cohort_distribution_per_hectare.csv
- plant_functional_type_species_classification_base.csv
- plant_functional_type_species_classification_maximum_height.csv

data/derived/plant/traits_data
- plant_stoichiometry.csv
- t_model_parameters.csv

Note: the files that are being tracked originally got added to the repo before we started using Globus, so perhaps I can just tell Git to untrack it? But I'm not sure if that's the correct way to handle this

arne-exe:
Nice, that worked @hrlai, I deleted them in #259

PR #241, opened 2026-03-19
Draft: FG population density tool

Baizurah:
Sharing this as a draft for review.

Tool 1 – Population Density

Input: animal_cohort_data.csv
Analysis: Calculates total population density (individuals/m²) for each functional group across timesteps
Output: Line graph showing population density over time, with one line per functional group
What the function does
This script calculates population density (individuals/m²) for each functional group over time using animal_cohort_data.csv. It groups individuals by timestep and functional group, divides by total simulation area, and produces a line plot of density trends. Would really appreciate any feedback so I can refine this further 🙏 Also sharing the figures produced using the herbivore only (Maliau_FG level 1) and the FG excel file from example data.

Figure_1_pop density testing function
Figure_2
@TaranRallings @nickwctan @annarallings


TaranRallings (commented):
Looks like you're off to a good start!

One of the things you want to make sure of is that you're being explicit about input/output typing and that you've got a full docstring in there. 

for example:

`def alpha_i_k(alpha_0_herb: float, mass: float) -> float:`

In this function, in the signature, we say that an input is `alpha_0_herb` and that its type is `float`. At the end of the signature, we use the `-> float` to specify that this function returns a float. 

    Args:
        alpha_0_herb: Effective rate per unit body mass at which a herbivore searches
          its environment in m2/(day*g).
        mass: The current body mass of the foraging herbivore in g.

    Returns:
        A float of the effective search rate in [m2/day].`

In the triple-quotes, we give the brief description, followed by the sections explaining the Args (arguments) and what it returns.

analysis/animal/fg_population_density.py
```diff
@@ -0,0 +1,38 @@
...
+    # group the functional group per each timestep
+    d = df.groupby(["time_index", "functional_group"])["individuals"].sum()
+    d = d.reset_index()
+    """Assumptions: The current simulation landscape area unit is in (m2) and need to be adjusted if grid size or cell size changes for each run. Current assumption is 50x50 grid cells with 100mx 100m per cell."""
+    # calculate simulation landscape area
+    cell_area = cell_size * cell_size
+    total_area = cell_area * n_cells_x * n_cells_y
+    # calculate population density
+    d["density"] = d["individuals"] / total_area
+    print(d.head())
```
davidorme: This is probably a debugging line, but again, if you leave it in the code then the function will output this every time a plot is generated. That is unlikely to be what you want.

davidorme (changes requested):
@Baizurah This is a great start and it's actually a really useful point from which to add feedback about the coding structure and Python stuff.

The comments and ideas are mostly just about how to structure Python modules (which you would use like `source("functions.R")` or `library(functions)` in R to add functionality). 

On what the function does, there are actually _three_ things bundled into a single function:

* Loading the data,
* Aggregating individuals counts to cell densities,
* Generating a density plot

If you think about how you might use this in a notebook, those are all separate steps and it is likely that it will be useful to keep them separate:

* You only want to load the data once (and you might want to do something to it before plotting - maybe filtering out some FGs?). Currently it is loaded and then discarded when you create the plot, and you can't interact with it first.
* You might want to do other things in the notebook with the aggregated densities but at the moment these are calculated inside this function and then lost when the function finishes.

So - loading is a simple thing - maybe you just use pandas directly in the notebook, but I think you have two functions here: `calculate_fg_densities` and `plot_fg_densities`. 

The other thing I'd add is we probably want to keep Python functions in a separate directory (`tools/python`) rather than having them mixed in with the other files. That's going to make it a lot easier to start thinking about packaging them up later on and to start thinking about adding some testing.

Lots to think about - do ping back here if you need more detail.

annarallings (changes requested):
Good start - please review and update with the suggestions, particularly on the metadata. It can be a work in progress in the repo but the metadata needs to be complete whether its R or python.

Baizurah:
Hi all,

Thank you so much for all the revisions, and suggestions. I really appreciate everyone’s time and input on this. Extra thanks to @davidorme for the detailed comments, they’re super helpful and I understand the requested changes.  I’ll work through the revisions and will ping here once I’m done.

Baizurah:
@TaranRallings @davidorme @annarallings When you have a chance, I’d appreciate any feedback or suggestions for further changes. Thank you!

nickwctan (commented):
May I suggest moving fg_population_density.py into tools/python/src/ve_data_tools to store all our helper functions in one place? These functions can be imported to be used in the analysis folders. See my comments in #516 and Lela's previous PR #503

Baizurah:
> May I suggest moving fg_population_density.py into tools/python/src/ve_data_tools to store all our helper functions in one place? These functions can be imported to be used in the analysis folders. See my comments in #516 and Lela's previous PR #503

I've drafted this before the tools folder created, so naturally it is in the outdated location now. I've consult @hrlai about moving it before resolving the issues, and I've decided that I will move once all issues/conversations here are resolved. @nickwctan

Baizurah:
@TaranRallings @davidorme Tagging for visibility. I'd appreciate feedbacks so i can close this soon.

annarallings (approved):
LGTM though as far as I understand it's not a traditional "tool" but an analytical script that can be reused on different outputs, which seems like it belongs in the analysis folder. (but if my thinking is incorrect, please do clarify) It would be great to see this used in a notebook environment on the next iteration.  I would also recommend going through and adding more comments on your code to talk through the steps. The notebook format would be helpful for this too. 

I am going to approve as it's just a py script for now which is fine and I know you're going to be continuing to rework it too.

analysis/animal/fg_population_density.py
```diff
@@ -0,0 +1,435 @@
...
+    if n_cells_observed > n_cells_expected:
+        raise ValueError(
+            f"The data contains {n_cells_observed} unique "
+            f"'{grid_cell_column}' values, but n_cells_x * n_cells_y = "
+            f"{n_cells_expected}. Check that n_cells_x and n_cells_y match "
+            "the simulation that produced this data."
+        )
+
+
+def parse_territory_cells(territory_value: object) -> set[int]:
```
davidorme: This is ultra picky but the typing could be more precise:

```suggestion
def parse_territory_cells(territory_value: str | list[int]) -> set[int]:
```

analysis/animal/fg_population_density.py
```diff
@@ -0,0 +1,435 @@
...
+
+    print(density_df.head())
+
+    if output_csv is not None:
+        density_df.to_csv(output_csv, index=False)
+        print(f"Density table saved to {output_csv}")
+
+
+if __name__ == "__main__":
+    main()
```
davidorme, line 435: It isn't clear what this section is for - see the main review comment.

davidorme (changes requested):
This is really nice, well done. It's very cleanly structured and commented and easy to review, so great work. 

My main comment is that it isn't clear why you've added a `main` function and made the script callable to run that function.  Your `main` function runs the functions in the module using a set of test data. That's great as a check that things are working - and I can completely see why you'd include something like this when developing the code as a quick way of running checks. But... now the code is complete, you'll likely never use that `main` function or run the module again - after all, it is running a very specific thing on a dummy test file.

So, I think you've got two choices:

1. You could delete it `main` and remove the `if __name__ == "__main__":` section. That's easiest.
2. Alternatively, you could turn the `main` code into a simple `pytest` of the functions in this module, which is really what it is. That's more complicated - you'd probably move the python module itself from the `analysis` directory into the `tools/python` and then add a small test file that imports those tools and runs the code currently in `main`.

davidorme:
> LGTM though as far as I understand it's not a traditional "tool" but an analytical script that can be reused on different outputs, which seems like it belongs in the analysis folder.

In practice, there is almost no difference between these things and it purely comes down to how it is easiest to structure and maintain the code. My reading is that the `fg_population_density.py‎` file is a module that provides a bunch of commonly used functionality - it's not a script that you'd run but part of a toolbox you'd import into scripts. From that, I'd argue that it goes into `tools/python` - the only real difference is that if you are importing functionality, it is easier to have only one location that you have to maintain as a local python module.

Baizurah:
I’ve moved fg_population_density.py into tools/python/src/ve_data_tools, removed the main() block, and also added pytest coverage for the landscape and territory density calculations. Ruff/pre-commit and the tests are passing. @davidorme when you have a chance, could you have another look?

PR #243, opened 2026-03-20
Animal - Added trophic flow analytical tool and test

nickwctan:
This PR adds an analytical tool to visualise the animal trophic interaction output from Virtual Ecosystem and calculates the trophic mass flow according to carbon, nitrogen or phosphorus.  

Currently, by default, it tracks the carbon sum of resource pool (AKA resource kind) of all consumer_cohort in each timestep. This outputs faceted plot for visualisation purposes. 

I envision that this will be extended in the future to create multiple graphical outputs to analyse trophic interactions. 

@davidorme Is there a way we can mimic the pytest pathing in Virtual Ecosystem so that it also works similarly in the ve_data_science repo? At the moment I have placed the test file in the same folder, which is not ideal.


davidorme (changes requested):
@nickwctan This is really neat - hadn't realised you were also building in testing.

I've got basically two types of comments:

1. You have added the `if __name__=="__main__":` sections to both files, but neither of those really reflect how the files should be used in practice (for different reasons for the tests vs the tool).
2. Your methods on the tool are using `return self` a lot. That isn't generally how class methods are used in Python - if your method is updating the class instance (i.e. doing `self.something = "a thing"` then the method can simply return `None` and the class instance that you ran the method on has been updated.

```python
In [2]: class Test:
   ...: 
   ...:     def __init__(self):
   ...: 
   ...:         self.a = 12
   ...: 
   ...:     def update_a(self) -> None:
   ...: 
   ...:         self.a = 24
   ...: 
   ...: 
   ...: test = Test()
   ...: test.a
   ...: 
Out[2]: 12

In [3]: test.update_a()
   ...: test.a
Out[3]: 24
```

ETA:

3. You have a ton of `print` statements. That is great for debugging - do you actually want these unavoidably appearing in all your notebooks that use the tool? You _can_ filter out printed messages in notebook code, but that isn't something users would normally expect to have to do. 

    You _could_ use the `logging` system to print these messages, which allows a user to suppress logging. I think you have to make a decision about whether these are debugging messages that were useful for you during coding (but should now just be deleted) or if they are actually a useful output that users might sometimes want (in which case, `logging`, but more complex!)

nickwctan:
> Don't return self here. The way to do this is return None and have the object just update its own internal state in place.

if we separate into multiple functions, the user would have to type out each of them out and might forget a step? I understand the idea behind that to keep it flexible, which would benefit when I would like to expand my class. 

If we keep the multiple function approach like you suggested, does it make sense to also have a lazier function to run the complete analysis pipeline in one call, something like:
```
    def run_analysis(
        self,
        convert_units: bool = True,
        group_data: bool = True,
        pivot_data: bool = True,
        plot_faceted: bool = False,
    ) -> None
```
The idea behind this is that (apart from us using it for validation) users can simply run a few lines to spit out a bunch of graphs for visualisation.

tools/python/animal/test_trophic_mass_flow.py
```diff
@@ -0,0 +1,85 @@
...
+    assert analysis.df is None
+    assert analysis.pivoted_df is None
+
+
+def test_load_data(analysis):
+    """Test load_data method."""
+    # call the method and store what it returns.
+    result = analysis.load_data()
+    # check method returns self (For method chaining)
+    assert result is analysis
```
davidorme: Should use `isinstance`?

tools/python/animal/test_trophic_mass_flow.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_process_data(analysis):
+    """Test process_data pipeline."""
+    # call load_data() method
+    analysis.load_data()
+    # calls process_data() running convert_units, group_and_aggregate(),pivot_data()
+    # store what method returns in "result" variable
+    result = analysis.process_data()
+    # checks if "result" is the same object as what we started
+    assert result is analysis
```
davidorme: `isinstance` again?

davidorme (changes requested):
You've still got a ton of `print` statements. You want reporting output to be under user control or your notebooks will get spammed with information. Your specific `summary` method is an elegant way to do that - the user can request a description when they want to. If you still want print output when methods are running then you should provide a verbose/quiet mode: you can wrap all the `print` statements in `if verbose:` but it probably is cleaner to use `logger` and mute output that way.

To give a practical example, have a look at this code - I've overdone it to show different levels but you only really need to use `LOGGER.info` to replace `print`.

```python
In [1]: import logging
   ...: 
   ...: 
   ...: # Get a logger and add a stream handler to write to stdout/stderr like print()
   ...: LOGGER = logging.getLogger(__name__)
   ...: LOGGER.addHandler(logging.StreamHandler())
   ...: 
   ...: def fussy_function(log_level: int | None = None):
   ...:     """This version allows individual methods/classes to be muted, but it seems wildly
   ...:     fussy."""
   ...: 
   ...:     if log_level is not None:
   ...:         current_level = LOGGER.level
   ...:         LOGGER.setLevel(log_level)
   ...: 
   ...:     LOGGER.warning("WARNING: I am here!")
   ...:     LOGGER.info("INFO: I am here!")
   ...:     LOGGER.debug("DEBUG: I am here!")
   ...: 
   ...:     if log_level is not None:
   ...:         LOGGER.setLevel(current_level)
   ...: 
   ...: 

In [2]: fussy_function(log_level=logging.CRITICAL)

In [3]: fussy_function() # Default
WARNING: I am here!

In [4]: fussy_function(log_level=logging.INFO)
WARNING: I am here!
INFO: I am here!

In [5]: fussy_function(log_level=logging.DEBUG)
WARNING: I am here!
INFO: I am here!
DEBUG: I am here!

In [6]: def function():
   ...:     """This version is more practical - users will typically want to set a logging level
   ...:     for an entire analysis/notebook."""
   ...:     LOGGER.warning("WARNING: I am here!")
   ...:     LOGGER.info("INFO: I am here!")
   ...:     LOGGER.debug("DEBUG: I am here!")
   ...: 

In [7]: LOGGER.setLevel(logging.CRITICAL)
   ...: function()
   ...: 

In [8]: LOGGER.setLevel(logging.DEBUG)
   ...: function()
WARNING: I am here!
INFO: I am here!
DEBUG: I am here!

In [9]: LOGGER.setLevel(logging.WARNING)
   ...: function()
WARNING: I am here!

In [10]: LOGGER.setLevel(logging.INFO)
    ...: function()
WARNING: I am here!
INFO: I am here!

In [11]: LOGGER.setLevel(logging.DEBUG)
    ...: function()
WARNING: I am here!
INFO: I am here!
DEBUG: I am here!

In [12]: 
```

tools/python/animal/trophic_mass_flow.py
```diff
@@ -0,0 +1,382 @@
...
+            Does not return but updates the dataframe.
+
+        """
+
+        groups = group_by or self.config["group_by"]
+        agg_col = agg_column or self.config["value_to_sum"]
+
+        print(f"Grouping by {groups} and summing {agg_col}...")
+
+        self.group_df = (
```
hrlai:
```suggestion
        self.grouped_df = (
```

Is this name a typo by any chance?

nickwctan: yes it was, thanks for picking that up!

davidorme (approved):
LGTM, @nickwctan - nice clean, well-structured Python code.

PR #244, opened 2026-03-24
180 remove downward shortwave radiation from plant input data

### LINKED ISSUE #180: Remove downward shortwave radiation from plant input data
`downward_shortwave_radiation` has been calculated by Lela and is now also part of the climate input data files.
So, when running the VE, adjust the config to use this climate data file instead of the plant input file for this variable.
This means that `downward_shortwave_radiation` can be removed from the plant input data and preparation scripts as it is no longer needed here.

arne-exe:
Fixes #180 by removing shortwave radiation and its time dimension from the input data scripts and related data files.
Also removes subcanopy vegetation and seedbank biomass from the plant constants.


arne-exe:
New plant constants files added to derived data folder and Maliau 1 scenario data folder.

hrlai (approved):
Looks good!

PR #248, opened 2026-03-26
247 Removed leaf_lignin from script and input data files

### LINKED ISSUE #247: Remove leaf_lignin from plant constants
As of the new latest release, `leaf_lignin` is no longer used (`senesced_leaf_lignin` still is though).
So, `leaf_lignin` should be removed from the plant constants input data file.

arne-exe:
Fixes #247 by removing leaf_lignin from script and input data files.


hrlai (approved):
Looks good! Will we also update the plant config file on Globus separately?

PR #259, opened 2026-04-03
258 Expanded PFT species classification

arne-exe:
Expands PFT species classification list (#258).


hrlai (approved):
Looks good and straightforward. Only thing (not crucial) is to consider having a master script that source all these files to reproduce output in one go, just like your commented pipeline above. Though I understand that this is not always straightforward!

PR #260, opened 2026-04-06
255 soil and litter cnp triplets

### LINKED ISSUE #255: Update soil and litter initial data to use CNP triplets
Now that [soil ](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1461)and [litter ](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1432)nutrients run on triplets, we should update the initial Maliau datasets to do the same.

hrlai:
This PR updates relevant soil and litter variables in the Maliau scenario to triplet arrays (and their associated metadata), so the input data are compatible with the recent code changes in https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1461 and https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1432

A lot of the diff are due to switching from the ncdf4 package to the RNetCDF package, as only the latter allows for character-vector dimensions (important for the C, N and P element new dimension).


hrlai:
@jacobcook1995 this is ready for review when https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1495 is merged. However I decided to keep the dimension in reverse order in R, so that it reads in the right order in VE. Just to be explicit, and more robust against future changes on the VE side in case we decided not to switch to `sel()`.

analysis/litter/initialisation/maliau_litter_data.R
```diff
@@ -281,59 +360,84 @@ litter_meta_df <-
...
-  vardef[[i]] <-
-    ncvar_def(
-      litter_meta_df$variable[i],
-      litter_meta_df$unit[i],
-      list(xdim, ydim)
-    )
+litter_vars <- litter_meta_df$variable
+for (i in litter_vars) {
+  if (str_detect(i, "_cnp")) {
+    var.def.nc(ncout, i, "NC_DOUBLE", rev(c("x", "y", "element")))
```
jacobcook1995, line 398: I think a comment should be placed somewhere around here, because at some point this won't be needed (and might actually break things if we decide to enforce a particular variable order)

analysis/soil/initialisation/maliau_soil_data.R
```diff
@@ -451,59 +474,84 @@ soil_meta_df <-
...
-# create netCDF file and put arrays
-ncout <- nc_create(ncfname, vardef, force_v4 = TRUE)
+# convert dataframe to arrays
+array_list <- vector("list", length(soil_vars))
+names(array_list) <- soil_vars
+for (i in soil_vars) {
+  if (str_detect(i, "_cnp_")) {
+    triplet_tmp <- do.call(rbind, dat[[i]])
+    array_list[[i]] <-
+      array(triplet_tmp, dim = rev(c(maliau$cell_nx, maliau$cell_ny, 3)))
```
jacobcook1995, line 534: Same comment as above about adding a note about why rev is needed here

jacobcook1995 (approved):
LGTM, just a couple of minor comments

hrlai:
Thanks @jacobcook1995 I have added a quick note to explain why `rev()` is needed.

PR #261, opened 2026-04-07
246 Scaled pft propagules to cell area used

### LINKED ISSUE #246: plant_pft_propagules calibration
The input data for `plant_pft_propagules` used the value 1000 reported by Metcalfe and Turner (1998; https://www.jstor.org/stable/2559870).
However, this value is the count of seeds in soil seedbank per m2. Currently, the input data gives 1000 seeds to each PFT, so need to update the script so that 1000 is multiplied by the cell area used in the simulation, and then divide this number by the amount of PFT's.

Another thing to consider is how to distribute these seeds across PFTs, should it be done evenly or do some PFTs have more seeds in the seed bank than others? The reference below may have some good info:
GARWOOD, N. C. 1989. Tropical soil seed banks: a review. Pp. 149-209 in Leck, M. A. & Parker, V. T. (eds). Ecology of soil seed banks. Acadenmic Pr-ess, San Diego (https://books.google.com.my/books?hl=nl&lr=&id=lXSmM24lt-cC&oi=fnd&pg=PP1&dq=GARWOOD,+N.+C.+1989.+Tropical+soil+seed+banks:+a+review.+Pp.+149-209+in+Leck,+M.+A.+%26+Parker,+V.+T.+(eds).+Ecology+of+soil+seed+banks.+Acadenmic+Pr-ess,+San+Diego.&ots=IFRMPBiDqR&sig=Q3e1cWuVNQ--xFa_A_MemmznZlE&redir_esc=y#v=onepage&q&f=false).

arne-exe:
Multiplied seeds per m2 by the cell area used (10000 m2), assuming equal distribution across PFTs.

As the reference is originally from pioneer seeds, using 1000 per m2 might be an overestimation for the non-pioneer PFTs, but currently it's the best/only data we've got.

### DIFF
--- a/analysis/plant/netcdf_plant_input_data/netcdf_plant_input_data_Maliau_50x50.R
+++ b/analysis/plant/netcdf_plant_input_data/netcdf_plant_input_data_Maliau_50x50.R
@@ -96,10 +96,14 @@ pft_index <- unique(cohort_distribution$plant_cohorts_pft)
 # analysis scripts
 
 # plant_pft_propagules: matrix of cell_id by pft (so 4 by 250)
-# Use fill value = 1000 using value reported in Metcalfe and Turner (1998;
+# Use fill value = 1000 m-2 using value reported in Metcalfe and Turner (1998;
 # https://www.jstor.org/stable/2559870)
+# Then scale this according to the cell area used (here 10000 m2)
+# Then divide this number by the amount of PFTs (assuming equal distribution)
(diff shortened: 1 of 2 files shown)


hrlai (approved):
Looks good. I have added a minor comment that does not need to be address now, just a suggestion for the future :)

annarallings (commented):
LGTM

PR #262, opened 2026-04-07
169 Added trees below 10 cm dbh to cohort distribution

### LINKED ISSUE #169: PFT cohort distribution assumption
Writing this issue down just to have a log.

If I'm not mistaken I believe the cohort distribution works as follows:
- Take Maliau census data
- Subset to species belonging to PFT classification
- Count individuals per PFT per DBH class, on a per hectare basis

An important assumption here is that species not part of the PFT classification (or trees without species identified)
are not included in the cohort distribution.
While the relative distribution of PFTs is fine, it is possible that we are underestimating stem density per hectare due to the above.

A solution to this would be to scale the cohort distribution to a predefined stem density per hectare (for example an average density representative for primary tropical rainforests in Sabah).

arne-exe:
Fixes #169 by using the value reported in Kenzo et al. (2015) for Balai Ringin site, then distributing these across 3 dbh classes following the size distribution in Lee et al. (2002), then spreading evenly across the 4 PFTs.

### DIFF
--- a/analysis/plant/plant_functional_type/plant_functional_type_cohort_distribution.R
+++ b/analysis/plant/plant_functional_type/plant_functional_type_cohort_distribution.R
@@ -289,91 +289,20 @@ data_taxa <- data_taxa[
 # Assign each tree into one of these DBH classes
 # The value for dbh represents the midpoint of the dbh class
 
-data_taxa$dbh <- NA
+# Use 2000 mm as expected max dbh
+max_dbh <- 2000
 
-data_taxa$dbh[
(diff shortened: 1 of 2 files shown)


hrlai (changes requested):
Looks good. Made a small suggestion that hopefully is easy to revise. 

I think the Kenzo + Lee approach to estimate <10-cm DBH stem density is reasonable. An alternative is to explore power-law scaling but that's another can of worm.

hrlai (approved):

PR #264, opened 2026-04-08
263 Refining pft propagules, annual recruitment probability, and seedling/sapling density in cohort distribution

### LINKED ISSUE #263: Refining pft propagules, annual recruitment probability, and seedling/sapling density in cohort distribution
**Refining the estimate of pft propagules**
In #261 I rescaled the value of 1000 seeds per m2 (based on Metcalfe and Turner, 1998) to the cell area used in the simulation.
However, the value 1000 is based on pioneer species and likely leads to an overestimation of the seedbank.
Unfortunately there isn't a lot of alternative data that can be used to define the initial number of propagules for non-pioneer PFTs.

As an alternative, it may be more accurate to focus on studies measuring seedling establishment, then convert this to an estimate of the seedbank by using the germination probability (e.g., 2.3% over two years reported by Kennedy, D. N., & Swaine, M. D., 1992). Preferably, at the very least we should distinguish pioneers and non-pioneers (as the pioneer seedbank in primary forest should relatively be much smaller compared to the other PFTs).
It is worth noting that Miyamoto et al. (2024) reported 0 pioneer recruits in primary forest for Maliau.
Studies on seedling establishment:
- Kuusipalo, J., Jafarsidik, Y., Ådjers, G., & Tuomela, K. (1996). Population dynamics of tree seedlings in a mixed dipterocarp rainforest before and after logging and crown liberation. Forest Ecology and Management, 81(1-3), 85-94.
- Miyamoto, K., Aiba, S. I., Aoyagi, R., & Nilus, R. (2024). Logging impacts on above-and belowground forest biomass and production in Bornean lowland forests.

**Refining annual recruitment probability**
Currently, the annual recruitment probability only uses the germination probability by Kennedy and Swaine (1992). However, it's quite a shortcut to go from seed directly to "a small tree". There are likely multiple rounds of filtering throughout the germination, seedling and sapling stages.
To go from seed to small tree, at the very least the recruitment probability should account for mortality in the seedling stage (e.g., 16% over 1.5 years by Kuusipalo et al., 1996). 
Note that currently recruitment probability is a constant across PFTs, but for pioneers specifically this study could be used to account for seedling survival: Howlett, B. E., & Davidson, D. W. (2003). Effects of seed availability, site conditions, and herbivory on pioneer recruitment after logging in Sabah, Malaysia.

**Refining seedling/sapling density in cohort distribution**
In #262 I added 5000 seedlings/saplings per hectare (based on Kenzo et al., 2015) to the cohort distribution and distributed these across 3 dbh classes (based on Lee et al., 2002). While doing so, I assumed equal distribution across the different PFTs. However, I think equal distribution may not be the most accurate, especially for pioneers (because these should have a much lower relative frequency compared to the other PFTs).
To account for this, I think it makes sense to use the findings (i.e., 0 pioneer recruits for Maliau) by Miyamoto et al. (2024) to only spread the seedlings/saplings across the non-pioneer PFTs.

arne-exe:
See #263 for details.


arne-exe:
plant_input_data_Maliau_50x50.nc and plant_input_data_Maliau_50x50_simplified.nc updated in Globus

arne-exe:
data/derived/plant/traits_data/t_model_parameters.csv updated in Globus

data/derived/plant/csv_plant_input_data/plant_constants_Maliau_50x50.csv updated in Globus
data/derived/plant/csv_plant_input_data/plant_constants_Maliau_50x50_simplified.csv updated in Globus

arne-exe:
data/derived/plant/plant_functional_type/plant_functional_type_cohort_distribution_Maliau_50x50.csv updated in Globus

data/derived/plant/csv_plant_input_data/plant_cohort_data_Maliau_50x50.csv updated in Globus
data/derived/plant/csv_plant_input_data/plant_cohort_data_Maliau_50x50_simplified.csv updated in Globus

arne-exe:
Hi @hrlai or @annarallings, could you please have a quick look at this one too?

analysis/plant/netcdf_plant_input_data/netcdf_plant_input_data_Maliau_50x50.R
```diff
@@ -96,15 +96,70 @@ pft_index <- unique(cohort_distribution$plant_cohorts_pft)
...
+seedbank <-
+  recruits_per_hectare_without_mortality / germination_rate # all seeds across PFTs # nolint
+
+# Distribute seedbank evenly across PFTs, assuming no seedbank for pioneers
+# Round down to avoid decimal seeds
+
+seedbank_emergent <- floor(seedbank / 3)
+seedbank_overstory <- floor(seedbank / 3)
+seedbank_understory <- floor(seedbank / 3)
+seedbank_pioneer <- 0
```
hrlai, line 156: I think zero recruit for pioneer is plausible, but it is based on a single 50x50m plot. If that plot is really mature and closed-canopy, then I find zero pioneer recruit (>5-cm DBH) believable although throughout our Maliau landscape some grids are going to have canopy gaps, and then there will be non-zero pioneer recruits >5-cm DBH. This does not break your line of reasoning here but I think it is worth noting down, especially if you consider for plant input data to vary across grids later. This is also for initialisation for hopefully the run will balance out this initial stage later.

Another very minor point is that "pioneer" in their paper is basically Macaranga spp. In their undisturbed forest there were two Mallotus that can be associated with disturbance. I don't think you need to overwrite their decision though, since PFT is subjective to some degree, and again this is only for initialisation :)

hrlai (approved):
I think broadly this looks good for the purpose of initialisation and best guess. Comments are only minor discussion points.

annarallings (approved):
LGTM really interesting approach to solving this issue, I'd be curious to see the analysis of the outputs.

analysis/plant/netcdf_plant_input_data/netcdf_plant_input_data_Maliau_50x50.R
arne-exe: Thanks @hrlai, that's a very good point. Canopy gaps will be much smaller than the cell area used so I guess to account for this within cells we'd have to increase the overall pioneer seedbank for a specific cell that is expected to have more canopy gaps. I guess this would also require the recruitment probability in these cells to be different, and preferably PFT specific (at the moment this is a constant in the model). Spatial variation across cells would also require the PFT definitions to be cell specific, which at the moment they aren't. I'll add a note based on your comment.

PR #265, opened 2026-04-09
Update Python version requirement to >=3.12

nickwctan:
@davidorme the previous pyproject file did not allow me to use python3.14, hence this PR.


nickwctan:
@davidorme can you have a quick look at this to see if it makes sense? Thanks!

davidorme (approved):
Sorry @nickwctan - slow on this. 

I tend to leave the upper limit in place to the most recent release. That does effectively say that we haven't tested with the most recent Python but within this project I think we're okay just leaving it open.

I have added a new argument to `ruff` in pre-commit that checks the Python syntax is up to date with the lowest supported version. That made some minor changes (`tomllib` is now a standard library)

nickwctan:
Sounds good @davidorme . However, there was a failed check on ruff that's preventing the merge.

Issue #254, opened 2026-03-30
Sensitivity analysis of soil and litter initial values


hrlai:
**Aim**: to quantify the sensitivity of final states $y$ to initial states $x$, when the parameters or constants $\theta$ are fixed. (I will leave varying $\theta$ for another time, soon)

The problem is characterised by:
- a multivariate input $x$ (which by the way is some $x(0)$ )
- a multivariate output $y$ (which by the way is some $x(t)$ )
- the need to sample the high-dimensional $x$ and thereby varying it *efficiently*
- the need for a sensitivity measure

After a round of independent literature search, I have basically come to the same set of options as @lelavathy:
- local sensitivity analyses (e.g., Morris) that consider inputs one-at-a-time (OAT) --- computationally cheap but ignores nonlinearity and non-additivity
- global sensitivity analyses (e.g., Sobol) that consider inputs simultaneously --- pros and cons are opposite to that of local analyses

The main hurdle against global analyses is computational cost, but there are shortcuts both [mathematically](https://doi.org/10.1016/S0010-4655(02)00280-1) and [programmatically](https://doi.org/10.18637/jss.v102.i05). If I have understood it correctly, then based on a set of assumptions we simply need _two_ input matrices $x$ and $x'$, and one output matrix $y = f(x)$. The important fact that $y$ only depends on $x$ means that we only need to run VE _once_ for each combination of $x$.

In practice a sketch workflow with the help of the R package [sensobol](https://www.jstatsoft.org/article/view/v102i05) looks like:
1. Generate $x$ and $x'$. $x$ is a $N \times k$ matrix, where $N$ is the number of quasi-random samples and $k$ is the number of (soil and litter) input variables
3. Rescale $x$. By default $x \in [0, 1]$ in sensobol, so an extra step is required to rescale $x$ to the range of the Maliau input dataset
4. Feed $x_{i = 1}$ to VE by running it on a _single_ grid, get $y_{i = 1}$
5. Repeat Step 3 for $N$ times
6. Feed $x$ (and under-the-hood $x'$) and $y$ to sensobol, get a sensitivity index (we will start with the first-order and total sensitivity)

I expect Step 4-5 to benefit tremendously from the [HPC](https://github.com/ImperialCollegeLondon/ve_data_science/pull/253) and we should get parallerised results within minutes, since each iteration (Step 4) completes within minutes.

hrlai:
[...]

PR #278, opened 2026-04-20
277 refactor soil litter netcdf generation code

### LINKED ISSUE #277: Refactor soil and litter netCDF generation code
Write the last portion of the scripts into a function, so it can be reused in #254

hrlai:
This PR cleaned up the scripts that generate soil and litter netCDFs from dataframe, mainly by moving huge chunks of manual codes to two functions `convert_df_to_nc()` and `convert_array_to_nc`. A secondary objective is so that these functionalities can be reused in subsequent sensitivity analyses (#254). There is also `subset_nc()`, which is a wrapper function around `convert_array_to_nc` to further clean up the codes; this function may also be useful for https://github.com/ImperialCollegeLondon/ve_data_science/issues/257#issuecomment-4222088399.

Each of the three functions above are stored in eponymous R scripts. (It is possible to store them in something like a `tools` folder as discussed in #243 and #241 but I'm storing them under soil because I am not 100% sure about it's generality to other modules yet, though `subset_nc()` and `convert_array_to_nc` are designed to work with other modules in mind).

Some metadata files and other codes are also restructured to facilitate this refactoring.

Asking anyone among @annarallings @arne-exe or @nickwctan to take a look and approve. Thanks!

@annarallings in particular, for now I have chosen to document the three function codes above in roxygen metadata, which is standard practice for writing R packages. Would you like me to also add our R metadata template on top? It feels a bit messy and redundant doing so, but there is no harm I guess. I wrote the roxygen-style metadata so these functions are forward compatible with an R package, should we choose to do a user-facing util package one day.


arne-exe (approved):
> This PR cleaned up the scripts that generate soil and litter netCDFs from dataframe, mainly by moving huge chunks of manual codes to two functions convert_df_to_nc() and convert_array_to_nc. A secondary objective is so that these functionalities can be reused in subsequent sensitivity analyses. There is also subset_nc(), which is a wrapper function around convert_array_to_nc to further clean up the codes.
>
My personal preference is to have all the code in 1 script (as it's easier to understand what's going on), but I can see how it's useful to avoid repetition and for sensitivity analysis.

> Each of the three functions above are stored in eponymous R scripts. (It is possible to store them in something like a tools folder as discussed in https://github.com/ImperialCollegeLondon/ve_data_science/pull/243 and https://github.com/ImperialCollegeLondon/ve_data_science/pull/241 but I'm storing them under soil because I am not 100% sure about it's generality to other modules yet, though subset_nc() and convert_array_to_nc are designed to work with other modules in mind). Some metadata files and other codes are also restructured to facilitate this refactoring.
>
Thanks for this. It's good to know this exists.

>@annarallings in particular, for now I have chosen to document the three function codes above in roxygen metadata, which is standard practice for writing R packages. Would you like me to also add our R metadata template on top? It feels a bit messy and redundant doing so, but there is no harm I guess. I wrote the roxygen-style metadata so these functions are forward compatible with an R package, should we choose to do a user-facing util package one day.
>
Won't comment on this as I'm not familiar with it.

annarallings:
Re: metadata, I think I generally err on the side of including it in case we build a tool that scrapes this metadata (which has been on the wish list for a little while). I know it's redundant for some aspects, but might come in handy later on. If there is no harm, that's my preference. Feel free to keep it pretty brief.

annarallings (approved):
LGTM (see comment on metadata)

nickwctan (approved):
The subset script looks neat now!  

> (It is possible to store them in something like a tools folder as discussed in https://github.com/ImperialCollegeLondon/ve_data_science/pull/243 and https://github.com/ImperialCollegeLondon/ve_data_science/pull/241 but I'm storing them under soil because I am not 100% sure about it's generality to other modules yet

I think its fine to add into a directory like` tools/R/soil `for example for now. Maybe it would be cool in the near future to have one script for R and python respectively to subset for all modules (except animals).

hrlai:
Thanks @annarallings I have added the usual metadata; agree that it would be useful if we get a tool to make use of them one day.

@nickwctan I will move these scripts to another folder in another specific PR, once we have all voted on the folder name.

Waiting for PR #287 to close before merging this.

Issue #237, opened 2026-03-18
Maliau netCDF files need consistent dimensions and attributes


hrlai:
Currently the Maliau soil, litter, elevation and climate `.nc` files have different attributes for the `x`, `y` and/or `time` dimensions. Plant does not input an `.nc` file yet but may do so in the future (right @arne-exe ?) This is not a breaking issue yet, because VE does not enforce their consistency. 

For instance:
- soil `x` and `y` defined units in [m] but the era5 (climate) input data leave them undefined. 
- soil variables have dimensions in the order [x,y] whereas climate's are in [time,y,x] (note the opposite ordering of x and y, and I am not sure if this matters, [it may not](https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/1054#issuecomment-4073913151))
- climate has coordinates defined, not soil doesn't

There might be other differences that I haven't noticed. 

I think we may need an equivalent of a common dimensions and attributes defined like the [example data generation scripts](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/develop/virtual_ecosystem/example_data/generation_scripts/common.py). We already have the [TOML file for Maliau](https://github.com/ImperialCollegeLondon/ve_data_science/blob/main/data/derived/site/maliau_grid_definition_100m.toml) that can be leveraged. 

I am not suggesting any change soon, but it feels like something worth discussing. I dug into this because I was writing [a script to extract a single cell](https://github.com/hrlai/ve.sandbox/blob/main/single_grid/extract_one_grid.R) for every input dataset (for faster test runs), but the inconsistent dimensions were making it difficult.

lelavathy:
> Personally, I think it makes sense to prepare the x and y coordinates for all spatial input data using b), as this streamlines input data obtained from different datasets. Input data files can always have extra variables (e.g., x_original and y_original) with dimensions x and y, to allow converting the streamlined x and y back to the original coordinates. I don't think the VE would track these original coordinates, but I guess they can easily be loaded in post-processing scripts.

I agree with this approach.
For my climate and elevation datasets, I’ve included additional variables for latitude and longitude to preserve the original coordinate system. This way, even though the input is streamlined to use x and y, the real-world spatial reference is still available for post-processing and interpretation.

hrlai:
The soil and litter dimensions have been address in #255 and https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/1497

The coordinate system is being updated to use original UTM in #271 

[...]

PR #280, opened 2026-04-20
Fix grid origin: use corner-based xoff/yoff instead of cell centres

lelavathy:
## Description

This PR corrects the definition of `xoff` and `yoff` in `maliau_site_definition.py` to ensure consistency between the grid definition and input NetCDF datasets.

Previously, `xoff` and `yoff` were set to the **centre of the first grid cell**, which is inconsistent with the VE grid definition where these values represent the **lower-left corner (origin) of the grid**.

## Changes

**Before (incorrect):**

```python
xoff = ll_x_utm50N + res / 2
yoff = ll_y_utm50N + res / 2
```

**After (correct):**

```python
xoff = ll_x_utm50N
yoff = ll_y_utm50N
```

## Reason

In VE, the grid is constructed from `[core.grid]`, where:

* `xoff`, `yoff` define the **grid origin (lower-left corner)**
* Input datasets (e.g. ERA5, elevation) should use **cell centre coordinates** (`+ res/2`)

Using centre-based offsets caused misalignment between the grid and NetCDF inputs, contributing to inconsistencies in dimensions and coordinate interpretation.

## Related Issue

Addresses the issue discussed in #237


arne-exe (approved):
Looks good Lela. One thing to consider is to add some clarification on how the script should be used across different grids. At the moment, it seems that `maliau_site_definition.py` is a "general" script that needs to be modified depending on which grid is used.
You could, for example, add a brief explanation on this and highlight which aspects of the script need to be changed in the future (for example when a user would want to generate the `.toml` file for a different grid).

lelavathy:
Closing this PR as anew script has been developed that supports multiple scenarios stored within a single TOML file. This will be submitted in a new PR.

Issue #283, opened 2026-04-21
Tool to generate ve_run config files


hrlai:
We have been generating `ve_run` config files by hand. The problems are:
1. The handmade config files are not pushed and synced because they are treated as data files (no big deal, there's Globus, but see below). Or, there is a csv generated from a script, but the csv data are transferred to the TOML by hand.
2. The changes are not tracked or documented
3. The config generation is not automated or semi-automated, which is cumbersome for projects that run lots of scenarios that permutes input data, or troubleshooting in general (e.g., #257 , #254 and #279)
4. Eventually we might write a [user tool](https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/649) for QoL anyways

I am going to draft some R functions to generate config templates. I need them for #279 and #254 anyways.

For now, they are only aimed for ourselves. In other words, lots of the default values will be specific to our Maliau scenarios. They are mostly to allow us to flexibly set data paths, cell coordinate systems etc. 

This also echoes in spirit with how we generate the Maliau metadata. Since we are now thinking about #281 and #280, I think it is a good time we extend the thinking to config files too.

davidorme:
I've talked about this a bit over in #243 and #241 but I think it would be good to create a new root level `scripts` directory where we can collect shared R and python functions into prototype packages. We _could_ even make them into packages, although I think the complexities of running with packages in development mode might be a step too far

I take @arne-exe's point elsewhere (https://github.com/ImperialCollegeLondon/ve_data_science/pull/278#pullrequestreview-4137644966)  about keeping everything in a single script, but identifying which bits of code can be re-used is basically the start of generating a python/R package for scenario development and I think we'd like to get this process started.

hrlai:
@nickwctan @Baizurah whenever we create a new scenario, you would edit the animal config file `animal_config.toml` by hand, right? 

@arne-exe do you also do the same? Your script would output a csv like `plant_constants_Maliau_50x50.csv`, after which you add them to `[plants.constants]` in `plant_config.toml` by hand?

nickwctan:
> [@nickwctan](https://github.com/nickwctan) [@Baizurah](https://github.com/Baizurah) whenever we create a new scenario, you would edit the animal config file `animal_config.toml` by hand, right?

yes only the path changes. 
```
[animal]
functional_group_definitions_path = '../data/animal_functional_groups_Maliau_level1.csv
```

[...]

PR #285, opened 2026-04-21
Refactor maliau_site_definition.py to support multi-scenario TOML workflow

lelavathy:
### Description

This PR replaces the existing `maliau_site_definition.py` implementation with a
multi-scenario grid generation workflow for the Maliau Basin.

Instead of generating a single TOML file per run, the updated script supports
managing multiple grid scenarios within a single TOML file, stored under:

```id="1z3m1k"
[Scenario.<name>]
```

Following feedback from @hrlai, the workflow has been redesigned to support
**multi-scenario storage within a single TOML file**, improving organisation and
making it easier to manage and compare different grid configurations.

Additionally, as raised by @arne-exe regarding which parameters should be edited
manually, this implementation adopts a **configuration-based approach** using the
`get_all_configs()` function.

Predefined scenario configurations (e.g., `maliau_1`, `maliau_2`) are defined
within this function, where each scenario specifies:

* grid dimensions (`cell_nx`, `cell_ny`)
* resolution (`res`)
* spatial extent (`bbox` in WGS84)

Users only need to select the desired scenario when running the script, rather
than modifying parameters directly in the code.

For future use (e.g., generating a new grid), users can simply add a new entry in
`get_all_configs()` without changing the core logic of the script. This improves
consistency, enhances usability, and reduces the risk of errors from manual edits.

---

### Key Changes

* Refactored script to support **multiple grid scenarios**
* Added predefined scenario configurations (`maliau_1`, `maliau_2`)
* Introduced configuration-driven workflow via `get_all_configs()`
* Added interactive scenario selection
* Implemented persistent TOML handling (read → update → rewrite)
* Added write-once protection to prevent accidental overwriting
* Ensured VE-compatible structure (`core.grid` block always written)

---

### ⚠️ Breaking Change

* Previous behavior: one TOML file per run
* New behavior: single TOML file containing multiple scenarios

This may require updates to workflows that depended on the old output format.

Good idea—separating the *new additions* makes it much clearer for reviewers.

Here’s a clean way to add a **dedicated section for newly added changes**:

---

### 🆕 Newly Added in This PR Update

Following recent updates, the PR now also includes:

* Added **scenario-based `core.timing` configuration**

  * Each scenario now defines its own temporal settings
  * Eliminates the need for manual time editing in climate scripts

* Integrated **`core.timing` into TOML output structure**

  * Written under `[Scenario.<name>.core.timing]`
  * Ensures consistency with VE configuration design

* Added **human-readable temporal metadata** in scenario headers

  * Example: `# Temporal: 2010-2020 (11 years)`
  * Improves readability and quick inspection of scenarios

* Removed redundant scenario-specific TOML files:

  * `maliau_grid_definition_100m_10x10.toml`
  * `maliau_grid_definition_100m_50x50.toml`
  * Consolidated into a single `maliau_grid_definition.toml`

---

### Related Work

* Replaces the previous implementation from #280 (now closed)


hrlai:
I was able to run the script (if I set the relative path right) `python analysis/site/maliau_site_definition.py`.

Do you normally run it from _inside_ the `analysis/site` folder?

lelavathy:
> Do you normally run it from _inside_ the `analysis/site` folder?

Yes, I normally run it from inside the analysis/site folder.

analysis/site/maliau_site_definition.py
```diff
@@ -1,110 +1,298 @@
...
+# CONFIGURATION
+# ============================================================
+# Define all available grid scenarios.
+#    Each scenario includes:
+#    - cell_nx, cell_ny : grid dimensions
+#    - res              : grid resolution (meters)
+#    - bbox             : bounding box (lat_min, lon_min, lat_max, lon_max)
+
+
+def get_all_configs():
```
hrlai, line 95: For testing, I was able to add a `maliau_3` scenario here, but it wasn't clear how I would add it. I figured it out that I should rerun this script in the terminal, and then type `maliau_3`, but there could be a clearer instruction in usage_notes? I have made a few suggestions in-line.

analysis/site/maliau_site_definition.py
```diff
@@ -1,110 +1,298 @@
...
-            xoff=ll_x_utm50N + res / 2,
-            yoff=ll_y_utm50N + res / 2,
+# ============================================================
+# CONFIGURATION
+# ============================================================
+# Define all available grid scenarios.
+#    Each scenario includes:
+#    - cell_nx, cell_ny : grid dimensions
+#    - res              : grid resolution (meters)
+#    - bbox             : bounding box (lat_min, lon_min, lat_max, lon_max)
```
hrlai:
```suggestion
#    - bbox             : bounding box in WGS84 (lat_min, lon_min, lat_max, lon_max)
```

analysis/site/maliau_site_definition.py
```diff
@@ -1,110 +1,298 @@
...
+    path: data/derived/site/maliau
+    description: Multi-scenario grid definition file for VE simulations
 
 package_dependencies:
   - pyproj
   - tomli_w
   - shapely
+  - tomllib
+
+usage_notes: |
```
hrlai, line 55:
```suggestion
usage_notes: |
  - Add scenario information to the list under `get_all_configs` in this script
  - In the terminal, run this script `python maliau_site_definition.py` from the root directory
  - Add or edit each scenario at a time interactively by typing, for example, `maliau_1`, `maliau_2` etc.
  - A TOML file will be created at the `output_path`
  - To update or add another scenario to the TOML file, repeat the steps above
  - Scenarios are stored under [Scenario.<name>]
```

analysis/site/maliau_site_definition.py
```diff
@@ -1,110 +1,298 @@
...
+    # Convert grid bounds back to WGS84
+    grid_bounds = box(ll_x, ll_y, ur_x, ur_y)
+    grid_bounds_wgs = transform(to_wgs.transform, grid_bounds)
+
+    # Compute grid cell centre coordinates
+    cx = [(ll_x + res / 2) + res * i for i in range(cell_nx)]  # X centres (eastings)
+    cy = [(ll_y + res / 2) + res * i for i in range(cell_ny)]  # Y centres (northings)
+
+    # Assemble final grid definition
+    return dict(
```
hrlai, line 198: We have been adding spatial grid information, but in #283 I have found it useful to also have `core.timing` information here. Do you think we should include timing info here as well? Open to suggestions.

analysis/site/maliau_site_definition.py
```diff
@@ -1,110 +1,298 @@
...
+## ============================================================
+# MAIN RUN FUNCTION
+# ============================================================
+# Generate and write a grid scenario to the TOML file.
+# Loads existing scenarios, prevents overwriting, and ensures
+# a complete VE-compatible TOML structure.
+def run(grid_name):
+    """Generate and save a grid scenario to the TOML file."""
+    # Output file path
+    output_path = "../../../data/derived/sites/maliau_grid_definition.toml"
```
hrlai:
```suggestion
    output_path = "data/derived/site/maliau_grid_definition.toml"
```

Suggesting the relative path to start from the root directory, and suggesting the user to run this script from root. I hope this isn't a difficult change to your work habit? Most of the scripts and functions in this repo is designed to run from root, and for a general user their IDE's working directory will probably be root too.

hrlai: Another (minor) suggestion is for `output_path` to be an argument that the user can input too, but that's targeting an external user. I think for now, all of us will only output this maliau site definition to the same folder.

data/derived/site/maliau/maliau_grid_definition_100m_10x10.toml
hrlai, line 1: If we are using only one TOML, then this 10x10 version (and the other 50x50 version) should be deleted?

hrlai (changes requested):
Made some minor suggestions in-line and hopefully is easy for you to go through. Mainly the output path may need some discussion with the team 😃

analysis/site/maliau_site_definition.py
lelavathy: thank you @hrlai for adding a more detailed notes

data/derived/site/maliau/maliau_grid_definition_100m_10x10.toml
lelavathy: I agree, will make the changes

analysis/site/maliau_site_definition.py
lelavathy: > Another (minor) suggestion is for `output_path` to be an argument that the user can input too, but that's targeting an external user. I think for now, all of us will only output this maliau site definition to the same folder.

Yes, that’s definitely possible. For now, we’ve kept it fixed as
output_path = "data/derived/site/maliau_grid_definition.toml"
since we’re all writing to the same location.

If needed in the future—especially for external users—I can easily update the script to accept output_path as a user-defined argument.

analysis/site/maliau_site_definition.py
lelavathy: > We have been adding spatial grid information, but in #283 I have found it useful to also have `core.timing` information here. Do you think we should include timing info here as well? Open to suggestions.

That’s a good point. At the moment, only the ERA5 workflow is using core.timing, and I’ve been manually editing the time settings directly in the climate processing script.

If we include core.timing in the TOML here, it would actually make things cleaner—I could then read the timing information directly from the same file for the climate processing as well, instead of hardcoding it.

hrlai: Thanks, having `core.timing` will greatly facilitate an upcoming function in #289

lelavathy:
@hrlai I’ve made the necessary changes. Could you please take a look and review? If everything looks good, please approve the PR.

hrlai (approved):
Thanks for the changes, looks good now. It would be great if you could also remove the individual 10x10 and 50x50, and then replace them with the consolidated nested TOML 👍🏼

PR #287, opened 2026-04-22
replace lintr with air formatter for R codes

### LINKED ISSUE #286: Replace lintr with air formatter for R codes
It seems that lintr / styler is being superseded by the [air formatter](https://tidyverse.org/blog/2025/02/air/), which is now built into VSCode and Positron (the IDE our team uses).

I have been testing our air for a month now, and it does feel much easy and faster to use than lintr / styler. 

For R users, the slowest choke point of pre-commit check has always been lintr / styler. And it does not seem to be keeping up with the diversity of R scripts. For example, lintr / styler causes a lot of unnecessary troubles for function scripts --- because we are not building the scripts as part of an R package, it complains a lot about missing global vars (likely due to dependencies) and it adds a lot of burden on us to maintain the function codes. 

I propose to replace lintr with air in pre-commit. 

One reason is to see if it relaxes linting for function codes, and another is to be compatible with the built-in formatter of VSCode and Positron. Sometimes air-formatted codes do not pass lintr! And this is making me go crazy.

I am going to make a draft PR to test air with pre-commit first. Let's see.

hrlai:
This PR replaces the lintr formatter with air for R codes.

@davidorme can I get your thoughts on this? Some notes can be found from the associated issue #286
.

I tested the change in the pre-commit hook in another PR #277 and the CI was indeed very fast now (thanks to Rust).


hrlai:
Ok another "complication" is that pre-commit also just reformatted a bunch of R scripts in this repo by @arne-exe . So adding you to the review. But let's hear from @davidorme first?

.pre-commit-config.yaml
```diff
@@ -5,35 +5,27 @@ repos:
...
+  - repo: https://github.com/posit-dev/air-pre-commit
+    # Air formatter for R codes
+    rev: 0.9.0
     hooks:
-    - id: style-files
-      args: [--style_pkg=styler, --style_fun=tidyverse_style]
-    - id: lintr
-    - id: parsable-R
-    - id: no-browser-statement
-    - id: no-debug-statement
```
davidorme, line 41: Does `air` also handle these tasks? I'm not sure I'm _that_ bothered if it doesn't but they are useful - particularly the parsable one.

analysis/litter/chemistry_and_turnover/combine_parameters.R
```diff
@@ -138,12 +137,24 @@ param_summary <-
 # incorporate the parameter uncertainty of r, but I'm only using its mean
 # estimate for simplicity here
 # nolint start
```
davidorme: It looks like `air` does not use `nolint` tags and instead uses `fmt`:

https://posit-dev.github.io/air/formatter.html#skip-comments

We should remove these outdated tags throughout the repo and either accept the new formatted result or use the new tags.

davidorme (changes requested):
I'm happy to replace `lintr` with a better faster tool. Honestly, I've never been too happy with the speed of the `lorenzwalthert/precommit` suite - hugely valuable but using R to run them is a big install issue and they aren't fast.

I like the results - it's similar to the `ruff` style for Python. There are some things that initially make people mad - like when it wraps 

```R
y <- c(1, ..., 100)
y <- c(
    1,
    ...,
    100
)
```

You just have to get over these things (or turn formatting off if they are wildly egregious) because ultimately the job of the formatter is to stop everyone making up their own formatting rules and just do one thing consistently.

Can you also add an `air.toml` configuration file as recommended here:

https://posit-dev.github.io/air/configuration.html#configuration-recommendations

hrlai:
> There are some things that initially make people mad 

Yeap they have been making me mad too, but choosing the lesser of two devils here...

> Can you also add an air.toml configuration file as recommended here:

Added the default TOML setting. I've been using it for a while without finding a need to change the settings, but it'll be there for us to modify later.

.pre-commit-config.yaml
hrlai: No it seems, so I have added the three below back:
```
      - id: parsable-R
      - id: no-browser-statement
      - id: no-debug-statement
  ```

The CI still feels pretty quick so let's leave them. I agree that running `parse()` is a good one.

analysis/litter/chemistry_and_turnover/combine_parameters.R
hrlai: Sure I'll go through them one by one.

hrlai:
Ok @arne-exe I've made changes with @davidorme 's comments and now I need your help to check through your files, and remove/replace the `nolint` tags.

There are lots of files but it can be done quickly. In the project directory in VScode, use Edit > Find in Files and type "nolint", then go through all the search results, remove/replace `nolint` with `fmt: skip` (see [here](https://posit-dev.github.io/air/formatter.html#skip-comments)), and then commit the changes. 

Once you've done that we'll request another review by @davidorme . Please feel free to approve this yourself too. Thanks!

davidorme:
I think in general my advice to the whole data team is to **not use `#fmt: skip`** unless the result is actively wrong rather than displeasing to your eye. Learning to accept "opinionated formatters" might just be a trend, but once you let it go 🎶 🎵 and just live with it, it does take away time spent fiddling with line breaks and spacing etc.

hrlai:
+1 to that. Actually I just removed all nolints because I also found the air formats alright (as it has been the last month). I suspect that'll be the case for you too @arne-exe so hopefully the review is a quick work for you.

arne-exe:
@hrlai I'll try to look at this asap, it might not be this week though.

arne-exe:
> I think in general my advice to the whole data team is to **not use `#fmt: skip`** unless the result is actively wrong rather than displeasing to your eye. Learning to accept "opinionated formatters" might just be a trend, but once you let it go 🎶 🎵 and just live with it, it does take away time spent fiddling with line breaks and spacing etc.
>
In general I don't mind adjusting the code to match the required format to pass pre-commit.
In the majority of cases where I used `# nolint`  it was because I couldn't find another way to pass the commit otherwise (i.e., pre-commit was saying "change this to x", but when I changed it to x pre-commit changed it back and kept giving me the same error). The other cases were where variable names were too long, but that was only in a handful of cases.

arne-exe:
@hrlai to confirm, the best way to do this is to pull the remote version of this branch to my local machine, then on that branch make the changes to my scripts, then commit the changes on this local branch and push to origin right?

hrlai:
> @hrlai to confirm, the best way to do this is to pull the remote version of this branch to my local machine, then on that branch make the changes to my scripts, then commit the changes on this local branch and push to origin right?

Yeap, I can see that you're doing that already now. Just checkout into this branch, make changes, commit and push. This makes us co-writing this PR / branch basically 😃

arne-exe:
All the commits above called "More nolint removal" are the ones that pre-commit detected in this PR, I'll now check if there are any left that need removing.

arne-exe:
@hrlai that should be all of them!

hrlai:
Thanks @arne-exe I can confirm that there is no more nolint tags anywhere by using the Find in Files function in my IDE.

@davidorme would you be able to review the changes and approve if they look okay? Thanks.

davidorme:
> @hrlai to confirm, the best way to do this is to pull the remote version of this branch to my local machine, then on that branch make the changes to my scripts, then commit the changes on this local branch and push to origin right?

~That's right, @arne-exe~

For some reason, had this issue open in my browser and it hadn't refreshed with all the stuff you have already done doing exactly this 😄

davidorme (approved):
I haven't gone through all the files - it's a good change and we've got the local config. Shouldn't have changed any of the actual functionality, so I'm happy to accept the result 😄

PR #289, opened 2026-04-22
283 tool to generate ve run config files

### LINKED ISSUE #283: Tool to generate ve_run config files

hrlai:
This PR contains three R functions to build VE config files using outputs in the `analysis` folder. The goal is to replace/minimise hand-editing of TOML files for scenarios. See #283 for more motivations.

The three functions are in the `tools` folder (I don't think we need `R` and `python` subfolders because the file extensions are self-explanatory):
- `tools/build_config_template.R` is simply an import of VE's config builder into R. It is currently not used. I included it from a previous experimentation for completeness.
- `tools/build_config.R` is the main function, the script documents the arguments etc. but it is easier to see it in action in `analysis/site/build_config_maliau.R`, which reproduces the `maliau_2` configs by code, not by hand.
- `tools/collect_data_paths.R` is a helper function for `build_config()`. I decided for it to be a separate function script because it can be used elsewhere and not designed to be strictly used only by `build_config()`. Currently it is very tailored to the workflow of the data team but can be generalised in the future.

Note: there are lots of other changed files, but they are related to the [branch](https://github.com/ImperialCollegeLondon/ve_data_science/pull/287) that switches to air formatter, which has been merged into this branch to make function codes pass pre-commit checks.


hrlai:
There has been a change in the design of this PR. Instead of creating a full list of config settings with `build_config_template()` and then edit some of them, I have decided to simply let the user to construct configs as named lists, and then parse + write them to TOML files with the new function `build_config()`. The outputs resembles our current individual config files more closely (in fact almost 100% identical). I have been testing these built configs on the `maliau_2` setting and got similar behaviour. 

The output TOML files are designed to be modular (i.e., one for each VE module) with https://github.com/ImperialCollegeLondon/ve_data_science/issues/283#issuecomment-4302671340 in mind.

hrlai:
First PR comment has been updated and this is now ready for review.

hrlai:
@arne-exe @nickwctan @lelavathy @Baizurah I'm including everyone from the data team here but am aware that not all of us are R users. Happy to discuss this during the next meeting to decide how to review this PR. I would like to get this merged asap because a few other branches depend on this.

arne-exe (approved):
This looks neat, great job.
One thing that would be good to clarify is how these different scripts relate to each other / build on top of each other.
I previously mentioned that it's harder to understand what's going on when working across multiple scripts (and when using tools / importing functions from other scripts).
Could you please help think about a good way to keep these things clear as the web of scripts grows larger and larger?

Some ideas:
- expand the description in each script
- create a readme doc that explains the flow/process in more details
- document how to use each tool on readthedocs (or somewhere else)

hrlai:
> Could you please help think about a good way to keep these things clear as the web of scripts grows larger and larger?

Would be good to hear others' opinion about this too.

Meanwhile, just adding in that building these tools into a package could solve these issues. To me, sourcing these functions raw is only slightly different from loading a package at the beginning of a script (in the python world the difference is a minimal as having an `__init__.py` ?). We trust those packages and do not ask questions about them, but we ask a lot more about the functions when they are not in a package. I have been thinking about simply adding the functions' source code in the `package_dependencies` section of each script's metadata, would that help tracking their dependencies? Once we have a tool that make use of these metadata, we could easily build a flow chart of them. 

This is something that I can add to a few of the new PRs 😃

davidorme:
> I have been thinking about simply adding the functions' source code in the `package_dependencies` section of each script's metadata, would that help tracking their dependencies?

You mean adding the paths to files being called sourced? Presumably not adding the actual source code - that would be huge duplication?

If building these new scripts into an R package is the best way of solving usage, then we should go for it. I think it would be easiest to start it in a new repository. It definitely should use the standard package building tools and QA, so I think it adds some overhead, but the question is whether the advantages outweigh that cost.

hrlai:
> You mean adding the paths to files being called sourced? Presumably not adding the actual source code - that would be huge duplication?

Yes, I mean something like the following in the metadata:
```
#| ...
#| package_dependencies:
#|     - R_package_1
#|     - R_package_2
#|     - path_to_source_code_that_only_contains_functions
#| ...
```
Does that still sound like duplication to you?

Regardless, I think building a package is the best way forward, but yes it will cost us time. I already have an [experimental package](https://github.com/hrlai/ve.utils) here, but setting up the files with `usethis` from scratch is fairly easy these days, so I'm happy to create a new repo under Imperial. For now my gut feeling is that the cost slightly outweighs the benefits.

davidorme:
> Does that still sound like duplication to you?

No - not at all. Sure, the path will be in the file, but - although we don't use this now - having the same info in easily machine readable metadata would be a good thing. I'd be tempted to make it into a new section though rather than conflate it with packages? `source_files:` maybe?

hrlai:
Thanks both, I'm gonna merge this now, and move the discussion about source-script documentation to another Issue 😄

PR #291, opened 2026-04-27
279 run time as a function of cell number

### LINKED ISSUE #279: Run time as a function of cell number
> I would not expect the simulation run time to change with cell area. That would be a useful thing to confirm. With the number of cells, note that you're increasing the number of cells on both sides so run speed will increase as a quadratic function of the grid size. It might be worth trying 2x10, 3x10 etc and see if that's roughly linear. If the run time is still seriously non linear even with linear steps in grid size then something odd is going on. 

 _Originally posted by @davidorme in [#257](https://github.com/ImperialCollegeLondon/ve_data_science/issues/257#issuecomment-4222088399)_

Now that we have #278 set up let's put it to use on this exercise (before we get native spatial subsetting in VE).

@davidorme it seems that we have been using "cell area" and "cell number" interchangeably in the other thread, but you had cell number in mind, right? If it was cell area then it's more of a config setting thing.

hrlai:
This PR examines the amount of total runtime and runtime per grid to understand the computational cost of VE. See the original motivation [here](https://github.com/ImperialCollegeLondon/ve_data_science/issues/279#issue-4292910900).

We could continue to explore but I'd like to close this PR with the results below to prevent myself from spending too much time on this topic 😅 . It is mostly for curiosity and a companion for the ongoing profiling effort; it does not really contribute to the data team's priority (other than helping me to understand the best way of setting up runs for many scenarios, e.g., in a sensitivity analysis).

Tagging / requesting review from people involved in earlier discussions / may be interested, but I hope one of you will just volunteer to comment/approve this PR 😄 Main scripts to look at are under `analysis/troubleshoot`.

## Key results

- Can be found in the untracked figure `data/derived/troubleshoot/runtime_per_cell/runtime.png`. Pasting it here: image
- Note that this is a revised version of the older figure [here](https://github.com/ImperialCollegeLondon/ve_data_science/issues/279#issuecomment-4312140962), which is no longer correct.
- _Total_ runtime is virtually linear with increasing grid number from 10, 20, 30, ... to 100.
- But the slope if <10, which means runtime _per grid_ increases more slowly than increase in grid number.
- But also runtime per grid is not that linear, probably due to the relatively large overheads in setting up the files etc. for small-grid-number scenarios
- In conclusion, the overheads do not grow exponentially when data size increases and in fact it is cheaper to run VE on larger datasets on a per-grid basis. This is not a huge savings though, given the range of the Y-axis on the right graph.

## Other learned things during analysis
- Do not parallelise `ve_run` for assessments like this because I suspect the parallelization overhead is messing up the runtime (this is why the [previous results](https://github.com/ImperialCollegeLondon/ve_data_science/issues/279#issuecomment-4312140962) is likely misleading). Or I have not parallelised them appropriately. Topic for another day!
- Run VE for the full time length rather than truncating it in debug mode. This ensures that the difference between scenarios are more stable and less prone to ephemeral background processes that temporarily fluctuate CPU performance.
- It is possible to programmatically modify the config files for each scenario, no need to hand edit them! 
- I coded functions to build configs in [another PR](https://github.com/ImperialCollegeLondon/ve_data_science/pull/289), but did not end up using them here because editing from the `maliau_2` configs is easier and makes the code more intuitive

## Additional notes:
- I ran VE with minimal output files by disabling animal and plant cohort data, and even initial, final and continuous data because they are not needed. This allows us to focus on the runtime of model processes.
- I placed the analysis scripts under a new folder `analysis/troubleshoot` because they do not fit under any module
- There are lots of other changed files, but they are related to the [branch](https://github.com/ImperialCollegeLondon/ve_data_science/pull/287) that switches to air formatter, which has been merged into this branch to make function codes pass pre-commit checks.


lelavathy (approved):
This is a clear and well-documented exploration—thanks for putting it together.

arne-exe (approved):
Looks good. Will have a closer look at these scripts in the coming week to fully understand the steps but it looks like there's some great functionality in there (like the config builder and the data alterations).

jacobcook1995 (commented):
This analysis looks good but I'm a bit confused why there's so many changes to other files (some of which just look like linting changes). Did this PR branch of from another (as yet unmerged) branch?

If it did you should merge those before merging this one so that you can see just the relevant changes that this PR introduces. (you can also set the merge target to the PR to the branch that you branched off of, and that will show just the relevant file changes, when that branch gets merged then the PR's merge target gets automatically updated to main)

hrlai:
> This analysis looks good but I'm a bit confused why there's so many changes to other files (some of which just look like linting changes). Did this PR branch of from another (as yet unmerged) branch?
> 
> If it did you should merge those before merging this one so that you can see just the relevant changes that this PR introduces. (you can also set the merge target to the PR to the branch that you branched off of, and that will show just the relevant file changes, when that branch gets merged then the PR's merge target gets automatically updated to main)

Yeap, I wrote a note about this in the last point under "Additional notes" in the PR message. Once PR #287 has merged into main, I'll merge main into this branch, and hopefully the changed files will reduce.

jacobcook1995:
> Yeap, I wrote a note about this in the last point under "Additional notes" in the PR message. Once PR #287 has merged into main, I'll merge main into this branch, and hopefully the changed files will reduce.

Ahh yes I should have read the notes rather than jumping straight to the code 🫠. I've changed the base for this PR, so that it now shows the relevant file changes

davidorme (approved):
LGTM - thanks for looking into this.

PR #293, opened 2026-04-27
290 add tools folder

### LINKED ISSUE #290: Add `tools` folder to root and move scripts into it
This is a follow up on #278 to move function scripts into the `tools` folder.

#289 is already using the `tools` folder.

#243 and #241 are also doing similarly, we should combine them.

Potentially also wrangle in #285

hrlai:
This PR moves some custom R functions into the `tools/R` folder, drafted since another PR #289 .

There have been some discussions in Issue #290 and I let the devil go by placing the scripts in `tools/R`, instead of `tools` 👼🏼 

This PR is only intended to tidy up the soil and litter scripts. If we decide to merge this PR, then feel free to follow up with your own PRs that reorganise other scripts.


arne-exe (approved):
Looks good!

hrlai:
Thanks @arne-exe for the quick turnaround. Rest of the team, I'm gonna merge this now, but feel free to continue the discussion on Issue #290 , and reopen it if needed.

nickwctan:
Sorry this totally slipped past me.  my bad. I understand why you want to avoid sub directories. 

How are you with just having one subdirectory for R and python in tools? the reason is that I have just created a folder for testing_data under tools. Does that make sense? I am open to other suggestions to make this work :)

hrlai:
The tools/R folder is already in place on the main branch from the recent merged PRs 😊

hrlai:
Oh if you mean you want something like tools/python/another_folder that's fine by me. In R packages we typically have all functions only under R/, but in python it seems that more nested folders are the norm since you import them like tools.python.another_folder...?

nickwctan:
yes exactly @hrlai. Great then lets stick with what we have at the moment.

PR #294, opened 2026-04-28
Refactor elevation preprocessing script

lelavathy:
## Description

This PR refactors the Maliau elevation processing workflow by reorganising script location and improving grid handling.

---

## Changes

* Moved script from:
  `analysis/abiotic/data_download/`
  → `analysis/abiotic/elevation_data/`

* Replaced previous implementation with a new modular workflow

---

## Key improvements

* Introduced **scenario-based grid configuration**

  * Reads from `maliau_grid_definition.toml`

* **Updated coordinate system handling**

  * Changed `(x, y)` from **distance relative to grid origin** → **real UTM Zone 50N coordinates**
  * Ensures direct spatial consistency with VE grids and geospatial datasets

* Refactored into modular functions:

  * `load_scenario()`
  * `process_elevation()`
  * `build_dataset()`
  * `run()`

* Supports **interactive scenario selection**

* Produces VE-ready dataset in `(x, y, elevation)` format using real-world coordinates

---

## 🔹 Notes

* No change to core resampling logic (still bilinear)
* Output structure remains VE-compatible
* Old script removed to avoid duplication

---


analysis/abiotic/elevation_data/maliau_elevation_data_processing_script.py
```diff
@@ -0,0 +1,383 @@
+"""
+---
+title: VE Elevation Data Preparation and Grid Reprojection for Maliau Basin
+
+description: |
+  This script prepares Shuttle Radar Topography Mission (SRTM) elevation data
+  for use in the Virtual Ecosystem (VE) model using a scenario-based grid
+  definition. Unlike the standard workflow that relies on a fixed site definition,
```
hrlai, line 8: If this script is able to process elevation data from anywhere, then is it worth removing the word "maliau" from the script's name, the title, and the function name?

But also see further comments below.

analysis/abiotic/elevation_data/maliau_elevation_data_processing_script.py
```diff
@@ -0,0 +1,383 @@
...
+output_dir = Path("data/derived/abiotic/elevation_data")
+output_dir.mkdir(parents=True, exist_ok=True)
+
+
+# ============================================================
+# LOAD SCENARIO
+# ============================================================
+
+
+def load_scenario(toml_path, scenario_name):
```
hrlai, line 141: Some of the defined functions in this script can be move to `tools` if you plan to reuse them elsewhere? 

I think `load_scenario`, `process_elevation` and maybe `build_dataset` can be moved, but if you choose to do so then they need to be ensured to be general. I think `load_scenario` is intrinsically general, whereas `process_elevation` may be less so (so keeping it here maybe).

`build_dataset` can be general, but currently for you only elevation takes `x` and `y`. A extra time dimension is required to generalise its use case to climate datasets.

hrlai (approved):
Looks good to me, happy to have this merged in the current stage. Feel free to make changes according to the suggestions in this PR, or another PR that makes the functions more general and then move them to `tools`.

hrlai:
Adding to my comment re: potentially generalising the functions, there is potential is making them also useful for stuff you did in PR #106 . Not a priority for this PR of course 😄

analysis/abiotic/elevation_data/maliau_elevation_data_processing_script.py
lelavathy: > If this script is able to process elevation data from anywhere, then is it worth removing the word "maliau" from the script's name, the title, and the function name?
> 
> But also see further comments below.

Good point on making the script more general by removing “maliau” from the name. However, the script still relies on reading a site-specific TOML file to define the scenario (e.g., grid extent, resolution, origin).

vgro: Good point. There’s a limit to how universally applicable it is, since it depends on a specific input dataset, others might use different elevation data with different variables or band names.

That said, for running multiple scenarios in one area, it still makes sense to me how you set it up. If you wanted to reuse it for something like Silwood, would you create a separate “Silwood” version, or make it more general so everything is defined via the TOML file? That way, Maliau-specific details could be removed and you could even rename it something broader, like “processing SRTM for VE.”

analysis/abiotic/elevation_data/maliau_elevation_data_processing_script.py
```diff
@@ -0,0 +1,383 @@
...
+  The reprojected SAFE Project DEM (UTM Zone 50N) is also available from:
+  https://zenodo.org/records/3490488
+
+  References: -
+  Farr, T. G., et al. (2007). The Shuttle Radar Topography Mission (SRTM).
+  Reviews of Geophysics, 45(2). https://doi.org/10.1029/2005RG000183
+
+  USGS (2017). Shuttle Radar Topography Mission (SRTM) 1 Arc-Second Global.
+  https://doi.org/10.5066/F7PR7TFT (Accessed: 2025-09-18)
+
```
vgro, line 50: I really like your description of the steps and data set, it's really clear and useful :-)

vgro (approved):
This looks good to me. I agree with previous comments on potentially generalizing the script instead of the Maliau focus and outsourcing tool functions, but they are not deal breakers.

annarallings (approved):
LGTM

PR #299, opened 2026-04-28
add a single function to run VE in R

### LINKED ISSUE #298: Explore `ve_run_cli` in R
I have been using `processx::run` in a personal R package [ve.utils](https://github.com/hrlai/ve.utils) to call `ve_run` from R. It might be more compatible with Python users in our team if I stick with `ve_run_cli` using `reticulate`.

I will test it out on the example data and the sensitivity analysis #254 .

hrlai:
This PR adds a single function `ve_run()` to run VE from R. It is simply an import from VE's `ve_run_cli`, and is part of the team's experimentation with calling VE directly from R and Python, so we can include `ve_run` in for-loops etc.

In the function's metadata, there is a section `@example` that applied it to the `maliau_2` scenario. Sadly it does not have a progress bar, which is in VE's `main.ve_run` but not `entry_point.ve_run_cli`, I think? To judge progress, I look at the file writing in the `out` folder.

Not that the latest dev version does not work with the example data for some reason. So if you want to test run it on example data, use the release version of VE.

### DIFF
--- a/tools/R/ve_run.R
+++ b/tools/R/ve_run.R
@@ -1,0 +1,32 @@
+#' Run the Virtual Ecosystem in R
+#'
+#' @param args A concatenated string of arguments. For example,
+#'   \code{ve_run --install-example /usr/abc} can be replicated by calling
+#'   \code{ve_run_cli(c('--install-example', '/usr/abc/'))}
+#' @param venv Optional path to a virtual environment where ve_run is installed.
+#'
+#' @returns An integer indicating success (0) or failure (1). Output and logs
+#'   are saved to the paths specified in `args`
+#'
+#' @examples
+#'   config_path <- "data/scenarios/maliau/maliau_2/config"
+#'   out_path <- "data/scenarios/maliau/maliau_2/out"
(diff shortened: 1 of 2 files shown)


tools/R/ve_run.R
```diff
@@ -0,0 +1,60 @@
...
+#'     paste0(out_path, "/logfile.log"),
+#'     "--config",
+#'     "core.debug.truncate_run_at_update=4"
+#'   )
+#'   ve_run(args)
+
+ve_run <- function(args, venv = NULL) {
+  if (!is.null(venv)) {
+    # use virtual environment if specified
+    reticulate::use_virtualenv(venv)
```
arne-exe, line 52: As mentioned in #300, where would this need to be specified?

arne-exe (approved):
Looks good, also looked at this in #300.

tools/R/ve_run.R
hrlai: It's probably easier if we all go through it during the next meeting 😄

PR #300, opened 2026-04-28
254 sensitivity analysis of soil and litter initial values

### LINKED ISSUE #254: Sensitivity analysis of soil and litter initial values

hrlai:
This PR is still a draft. It adds a sensitivity analysis of the soil and litter initial values. More details to come.

Currently I have set up the pipeline to build input data and configs for multiple VE runs. There are 1000s - 10000s of runs required. I have tested the pipeline on only 5 permutations. I need to wait for the HPC to fully scale it up.


analysis/soil/sensitivity/input_data/build_data_sensitivity.R
```diff
@@ -0,0 +1,226 @@
...
+    plant_cohorts_dbh = mean(plant_cohorts_dbh),
+  ) |>
+  # reset cell_id to start from zero here too
+  mutate(plant_cohorts_cell_id = 0)
+write_csv(
+  plant_cohort,
+  paste0(out_dir, "plant_cohort_data_maliau_50x50.csv")
+)
+
+# Copy over other data that do not need to be modified
```
arne-exe, line 392: What determines if data needs to be modified or not, and why are these 2 files treated differently?

analysis/soil/sensitivity/input_data/build_config_sensitivity.R
```diff
@@ -0,0 +1,112 @@
...
+
+# input data paths for core.data.variable
+# first list the soil and litter data files to loop over
+data_paths_soil_litter <- paste0(
+  "../../data/",
+  list.files(
+    "data/scenarios/sensitivity_soil_litter/data",
+    "soil_litter_data",
+  )
+)
```
arne-exe: I assume the files in this folder are generated by one of the other scripts in this PR?
Perhaps specify which script is used to generate these.

analysis/soil/sensitivity/input_data/build_config_sensitivity.R
```diff
@@ -0,0 +1,112 @@
...
+        data = list(variable = data_path)
+      )
+    }
+  )
+
+# plants config
+plants <- list(
+  cohort_data_path = "../../data/plant_cohort_data_Maliau_50x50.csv",
+  pft_definitions_path = "../../data/plant_pft_definitions_Maliau_50x50.csv",
+  constants = as.list(plants_contants)
```
arne-exe, line 72: Here you're loading the constants with the aim to convert them to toml format instead of csv right?

arne-exe: Perhaps you can add a description at the top of this script to explain/summarize what it does.
In short, it looks like this script reads in the initial data that has been modified and then generates the config files for these?

analysis/soil/sensitivity/input_data/build_config_sensitivity.R
```diff
@@ -0,0 +1,112 @@
...
+      "litter"
+    ),
+    core = x,
+    plants = plants,
+    animal = animal,
+    path = config_dir
+  )
+  # Copy over soil microbial config that does not change across scenarios
+  file.copy(
+    "data/scenarios/maliau/maliau_1/config/soil_microbial_groups.toml",
```
arne-exe: >Copy over soil microbial config that does not change across scenarios
>
For someone else aiming to use this script, it suggests they would not need to adjust this, ever. If this is the case then there's no issue with always using the Maliau_1 file.

Also, depending on how you intend to use this script you could write it (particularly the paths) in a more generic way. Currently the script title is generic, implying it is more like a tool than a Maliau_1 specific script.

analysis/soil/sensitivity/input_data/build_data_sensitivity.R
```diff
@@ -0,0 +1,226 @@
...
+  ) |>
+  # paste variable names with their extra dimension names to treat them as
+  # individual variables
+  mutate(
+    variable = ifelse(is.na(element), L1, paste(L1, element, sep = "_"))
+  ) |>
+  select(variable, min, max)
+
+
+# Set up Sobol matrix -----------------------------------------------------
```
arne-exe, line 149: Please add a brief description on this Sobol matrix, perhaps also provide a reference for those interested in better understanding why/how this works.

analysis/soil/sensitivity/input_data/build_data_sensitivity.R
```diff
@@ -0,0 +1,226 @@
...
+n_sample <- 5
+mat <- sobol_matrices(
+  N = n_sample,
+  params = maliau_vars_init,
+  order = "first",
+  type = "QRN",
+  seed = 777
+)
+
+# rescale to Maliau ranges
```
arne-exe, line 174: Why is rescaling required?

analysis/soil/sensitivity/input_data/build_data_sensitivity.R
```diff
@@ -0,0 +1,226 @@
...
+        }
+      }) |>
+      convert_array_to_nc(
+        filename = paste0(out_dir, "soil_litter_data_", idx, ".nc")
+      )
+  })
+
+
+# Generate mean arrays ---------------------------------------------------
+# For other modules, fix their values at spatial averages
```
arne-exe: Please explain why you do this. And what average is this referring to, all cells across the entire grid?

analysis/soil/sensitivity/input_data/build_data_sensitivity.R
arne-exe: Probably within the context of spatial averages?

analysis/soil/sensitivity/input_data/build_data_sensitivity.R
```diff
@@ -0,0 +1,226 @@
...
+plant_cohort <-
+  read_csv(
+    paste0(in_dir, "plant_cohort_data_maliau_50x50.csv"),
+    show_col_types = FALSE
+  ) |>
+  group_by(plant_cohorts_pft) |>
+  # take average cohort size and DBH and floor cohort size to get integer
+  summarise(
+    plant_cohorts_n = floor(mean(plant_cohorts_n)),
+    plant_cohorts_dbh = mean(plant_cohorts_dbh),
```
arne-exe, line 383: You could focus on trees >10cm dbh, because the cohorts smaller than 10 cm dbh have a much greater number of individuals. Not sure how important ecological realism is in this sensitivity analysis? Also see comment on defining the "right" spatial average.

analysis/soil/sensitivity/input_data/build_data_sensitivity.R
arne-exe: Also, how does the selected spatial average affect the outcome of the sensitivity analysis?
How critical is defining the "right" spatial average?

tools/R/ve_run.R
```diff
@@ -0,0 +1,60 @@
...
+#'     paste0(out_path, "/logfile.log"),
+#'     "--config",
+#'     "core.debug.truncate_run_at_update=4"
+#'   )
+#'   ve_run(args)
+
+ve_run <- function(args, venv = NULL) {
+  if (!is.null(venv)) {
+    # use virtual environment if specified
+    reticulate::use_virtualenv(venv)
```
arne-exe: Where would this need to be specified? 

Side question, can you please explain how R runs the VE within a specific virtual environment. I'm not familiar with it (only VSCode) but would like to understand this.

arne-exe (approved):
Looks good @hrlai, a few comments mostly about providing a bit more context about why you do certain things.

hrlai:
Thanks @arne-exe for the early review, I'll come back to them when this is more shaped-up. Currently the parts are still subjected to changes; I just wanted a draft PR to remind myself. 😃

tools/R/ve_run.R
hrlai: It's probably easier if we all go through it during the next meeting 😄

PR #303, opened 2026-04-29
301 helper function for plotting

### LINKED ISSUE #301: Helper function for visualising VE netCDF outputs
Previously I've experimented with very crude plotting functions [here](https://github.com/hrlai/ve.sandbox/tree/main/code). Time to revamp them now that `tools/R/get_all_variables.R` and `analysis/soil/sensitivity/cell_id_to_xy.R` are available. (The latter will be moved into `tools/R` by PR #300)

I only have a rough fuzzy idea of how to design it around arrays with different extra dimensions beyond `x`, `y`, and `cell_id`, so suggestions are welcome.

hrlai:
This PR adds a function in `tools/R/tidy_continuous_data.R` and example usage in `analysis/soil/visualisation/visualise_continuous_data.R`. The function `tidy_continuous_data()` takes
1. the all_continuous_data netCDF output data
2. optionally the initial_state netCDF
3. and a vector of state variable names

to tidy the state variables into a long-format dataframe, which should be highly compatible with, but not limited to, the visualisation grammar of `ggplot2`. The function also works across all modules (except for animal, which does not output netCDF, of course; that is being addressed in #241 ).


arne-exe (approved):
Nicely done, I've been manually writing out the variable extraction and conversion to long format but this looks more convenient, will try this tool out soon!

hrlai:
Thanks. Let us know if you manage to break it. We can fix it together. I haven't tested it exhaustively on every single state variable.

PR #315, opened 2026-05-07
Notes of validation

hrlai:
This PR puts some thoughts down on validation. @jacobcook1995 would you be able to take a look, since most of it are motivated by our ongoing conversions. The compiled html is on [Dropbox](https://www.dropbox.com/scl/fi/v3klnehzmt1k3bq56obtn/conceptual_blocks.html?rlkey=vgzka74geojbalp0cxz3njegu&dl=0) if you prefer. Feel free to add your thoughts. I hope to send this around very soon because the validation meeting is next week, thanks!

### DIFF
--- a/analysis/soil/validation/conceptual_blocks.qmd
+++ b/analysis/soil/validation/conceptual_blocks.qmd
@@ -15,7 +15,7 @@ format:
 
 - I learned about "validation" from the statistical modelling perspective; there may be some nuanced difference from process-based or mechanistic modelling.
 - I try to move across modules but apologies in advance if I have gotten anything wrong.
-- I try to think about validation generally, instead of only for Maliau itself. Solving this for Maliau = solving for any subsequent scenarios.
+- The data team members added lots of useful and contrasting perspectives while reviewing this document. I have included a selection of them in boxes.
 
 # Conceptual blocks
 
@@ -32,12 +32,67 @@ As a counter example, the soil input data had to be sourced from different study
 
 Even if abiotic inputs and outputs have a clear time mapping, the abiotic validation targets are affected by variables from other modules that do not (e.g., microclimate depends on plants). Because there is no spin-up phase in VE (except animals maybe), during the early time steps of simulation the state variables are likely doing *both* endogenic/internal stabilisation and reacting to exogenous factors. As such, even if we have clear validation targets, it is still hard to disentangle the effect of poor initialisation vs. poor parameter calibration.
 
+::: {.callout-note}
+## Team comments:
+Bai: 
+
+- Animal validation data are also unlikely to match the exact simulation start and end dates. 
+- I'm also wondering whether there are criteria or indicators we should use to identify when the system is sufficiently stabilised for interpretation or validation.
+
+Nick:
+
+- Do we validate our animal population across time, or just validate a snapshot of the population. 
+- Can we find empirical datasets of population or biomass fluctuations in function of time?
+
+Lela:
+
+- The modules have different response timescales. 
+- Abiotic variables such as temperature, hydrology, or microclimate can respond relatively quickly to short-term climate forcing, whereas variables such as soil carbon pools, nutrient dynamics, or vegetation structure may evolve much more slowly over time. 
+- Therefore validation/calibration approaches that are appropriate for fast abiotic processes may not necessarily work well for slower ecological processes, especially in the absence of a dedicated spin-up or equilibration phase.
+
+Jacob:
+
+- 100% on response timescales, this is basically why doing some kind of spin-up is reasonable for hydrology and microclimate, but not for soil where it can take 100s of years to reach a stable density of soil carbon (which is a real ecological effect, rather than a model artefact).
+
+:::
+
 ## Validate on Maliau or outside of Maliau?
 
 We have not decided whether the validation dataset should be from Maliau or not. If we initialise with Maliau or Maliau-like input data, then it makes sense to validation on another test dataset from Maliau. With the exception of abiotic, however, many input data are not exactly from Maliau, but similar forests. So it is probably okay to validate on non-Maliau datasets, plus there is often limited data for many of the state variables at Maliau.
 
 That said, if we validate soil on external dataset A and animal on external dataset B, and find that soil validates well but animal poorly, what does that mean? We cannot really conclude that soil is doing better than animal, because who knows if animal actually validates well on dataset A? Should we then all validate on the same dataset, which has a near-zero probability of having sampled all of our state variables? If it exists, we would have used it for input data already...
-[Nick has some thoughts on using global data for validation, i.e., if we validate the Maliau scenario onto global test dataset, and then redo for the SAFE scenario, then what does it mean to use the same global dataset twice?]{.aside}
+
+::: {.callout-note}
+## Team comments:
+Arne:
+
+- I think the goal for the Maliau scenario is to model the ecosystem (i.e., intact primary), rather than trying to validate Maliau exactly.
+- Perhaps it helps to separate whether we're validating a model-specific process, or the overall VE simulation as a whole. 
+- For model-specific processes, I think it's fine to use different datasets (i.e., your example of datasets A and B), because the goal here (I think) should be to calibrate the process/science within the specific model, we don't necessarily need to focus here on how the interaction with other models.
+- For validating the VE simulation as a whole, these interactions become more important.
+
+Hao Ran (follow up):
+
+- I think interactions are still important when we validating or calibrating a single process.
+- I think interactions are only unimportant if a global sensitivity analysis tells us so.
+
+Lela:
+
+- There is no long-term station dataset available from Maliau specifically for validating the abiotic or hydrology modules. Because of that, I was thinking that using datasets from VJR and Danum could be a reasonable alternative, since those sites are also relatively intact tropical forests.
+- This would also align with the idea that the Maliau scenario is intended to represent a broader intact Bornean rainforest system rather than an exact replication of Maliau itself.
+
+Bai:
+
+- No two primary/intact forests are ever truly the same. 
+- Perhaps the first step is defining the validation target itself, then identifying which ecological factors are most likely to influence that target, and only after that deciding whether an external dataset such as Danum is appropriate.
+
+Nick: 
+
+- So we have a concensus of Maliau as a representation of intact forests. 
+- However, if we then validate Danum, a similar intact tropical forest, how much would our validation datasets differ? I guess not so much.
+- Likewise, if we validate the Maliau scenario onto global test dataset, and then redo for the SAFE scenario, then what does it mean to use the same global dataset twice?
+
+:::
 
 ## Validate within or across modules?
 
@@ -47,6 +102,25 @@ Currently we are validating within modules, but we are also increasingly aware o
 2. Validation within module overlooks the possibility that, for example, a poorly validated plant variable could be due to a poorly initialised or calibrated soil variable.
 3. Validating across modules helps modules with hard-to-define targets. For example, plant validation datasets are relatively available but not animals, so why don't we just let animal use plant as validation targets? This is meaningful because if animal processes such as herbivory works well then we would expect the plant validation targets to do well too.
 
+::: {.callout-note}
+# Team comments:
+
+On the holistic goal of VE :
+
+- Arne: Yes and no. Validation within models is necessary to guarantee the overall science within each model (without this you would assume that their theoretical frameworks are complete and without flaws, it may be but we should test this).
+- Jacob: It's pretty hard to validate within a single model though, e.g. at the moment I can run the simulation and look at whether the soil results look plausible. You can pick out clearly incorrect behaviour (e.g. negative carbon pools), but "validating" the outputs really isn't possible because the subcanopy plants currently grow to form an impenetrable layer that block all light causing the soil to freeze. It is easier with some models (e.g. hydrology) because they are less linked to other models, but the biotic models are pretty tightly coupled.
+- Lela: Initially, I thought the abiotic/hydrology modules could be calibrated and validated more independently because they are more directly linked to physical processes and historical climate data. However, once the modules are coupled together, it becomes much harder to tell where unrealistic behaviour is actually coming from.
+- Nick: I think it would be across module for animals.
+- Bai: I think this works both ways. Within-module validation is still important to check whether specific processes are functioning sensibly, while across-module validation is needed to understand how those processes interact within the wider ecosystem
+- Hao Ran (follow up): My personal opinion is that validating within module does not guarantee the science within each model because the science of each model depends on other processes. As an example, even for a barebone resource--consumer food chain / food web model, the consumer dynamics still depends on the resource processes (e.g., basal plant growth), whether they are implicitly or explicitly modelled. When we isolate a module, the model becomes more and more phenomenological and hence departs further from the process-based design of VE.
+
+On the risk of sharing validation targets:
+
+- Arne: Animal validation may "pass" when using plant validation datasets, without knowing why (think of it like a black box). This type of validation will likely leave gaps / unvalidated processes in the animal model (to use your example/analogy).
+- Bai: It may also create confusion because a "good" plant result does not automatically mean the animal processes underneath are correct. For example, herbivore density might be too high, but feeding rate too low, and the plant outputs could still appear reasonable overall.
+
+:::
+
 # Another solution: focus on qualitative or emergent properties
 
 Some of the text above contains potential solutions; here's another that Jacob and I have discussed (with soil in mind, mostly).
@@ -55,3 +129,23 @@ Even if we managed to find a soil validation dataset, we feel that overly focusi
 [The soil nutrient pools now are in triplets, e.g., `soil_cnp_pool_maom`, but I'm just using carbon as an example here, hence `soil_c_pool_maom`.]{.aside}
 Do we then revisit `soil_c_pool_pom` only to get it right? Probably not, because (1) it also depends on `soil_c_pool_maom` and (2) the input data source probably contains both variables anyways. Most often than not we care about the whole carbon pool being partitioned correctly, rather than each pool hitting their own targets (if it happens, doesn't it just mean the former). In other words, since we will often be looking at groups of variables, then a validation target will often correspond to a group of variables. This naturally led us to think more about qualitative or emergent properties, e.g., that the ranking between soil nutrient pools should be preserved, soil fungi and bacteria should coexist, soil--microbe--plant stoichiometry should be preserved etc. And as you see it also naturally led us to think across modules.
 [There are other higher-level emergent properties that are not explicitly captured as state variables in VE, such as thinning rules, trophic distributions, network properties, energy compartments, biodiversity--productivity relationships etc., but they go beyond the discussion here.]{.aside}
+
+
+::: {.callout-note}
+# Team comments:
+
+Jacob:
+
+- I think the most relevant emergent property for soil is the carbon sequestration trajectories (i.e. is carbon being lost or stored by the soil over the time period in the manner the literature would lead you to expect).
+
+Arne:
+
+- What comes to mind here is to find a way to show which areas are calibrated well, and which ones are less.
+- What we do need to be cautious of is to avoid giving people a chance to invalidate our "emergent conclusions" because we haven't sufficiently shown/proven the within model functioning (see my other comment above).
+
+Nick:
+
+- At the moment for animals, I would say the collection of validation datasets and its ideas have been on emergent properties. They could be population turnover, growth rate, time to maturity, movement (which might be ambitious), trophic mass flow... 
+- I am currently looking at building a trophic network to potentially explore network properties and energy flow.
+
+:::
--- a/analysis/soil/validation/conceptual_blocks.qmd
+++ b/analysis/soil/validation/conceptual_blocks.qmd
@@ -34,20 +34,20 @@ Even if abiotic inputs and outputs have a clear time mapping, the abiotic valida
 
 ::: {.callout-note}
 ## Team comments:
-Bai: 
+Bai:
 
-- Animal validation data are also unlikely to match the exact simulation start and end dates. 
+- Animal validation data are also unlikely to match the exact simulation start and end dates.
 - I'm also wondering whether there are criteria or indicators we should use to identify when the system is sufficiently stabilised for interpretation or validation.
 
 Nick:
 
-- Do we validate our animal population across time, or just validate a snapshot of the population. 
+- Do we validate our animal population across time, or just validate a snapshot of the population.
 - Can we find empirical datasets of population or biomass fluctuations in function of time?
 
 Lela:
 
-- The modules have different response timescales. 
-- Abiotic variables such as temperature, hydrology, or microclimate can respond relatively quickly to short-term climate forcing, whereas variables such as soil carbon pools, nutrient dynamics, or vegetation structure may evolve much more slowly over time. 
+- The modules have different response timescales.
+- Abiotic variables such as temperature, hydrology, or microclimate can respond relatively quickly to short-term climate forcing, whereas variables such as soil carbon pools, nutrient dynamics, or vegetation structure may evolve much more slowly over time.
 - Therefore validation/calibration approaches that are appropriate for fast abiotic processes may not necessarily work well for slower ecological processes, especially in the absence of a dedicated spin-up or equilibration phase.
 
 Jacob:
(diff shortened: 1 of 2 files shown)


analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+
+Even if abiotic inputs and outputs have a clear time mapping, the abiotic validation targets are affected by variables from other modules that do not (e.g., microclimate depends on plants). Because there is no spin-up phase in VE (except animals maybe), during the early time steps of simulation the state variables are likely doing *both* endogenic/internal stabilisation and reacting to exogenous factors. As such, even if we have clear validation targets, it is still hard to disentangle the effect of poor initialisation vs. poor parameter calibration.
+
+## Validate on Maliau or outside of Maliau?
+
+We have not decided whether the validation dataset should be from Maliau or not. If we initialise with Maliau or Maliau-like input data, then it makes sense to validation on another test dataset from Maliau. With the exception of abiotic, however, many input data are not exactly from Maliau, but similar forests. So it is probably okay to validate on non-Maliau datasets, plus there is often limited data for many of the state variables at Maliau.
```
arne-exe, line 61: I agree. Not all input data is truly Maliau specific, although most should represent a similar study system (i.e., intact primary rainforest in Borneo). Keeping this in mind, validating Maliau specific input data with Maliau specific validation data is just not possible in many cases. Sure you can split up part of a Maliau dataset but I'm not entirely sure whether that provides better insights compared to using e.g. Maliau input and Danum validation. Also, I think the goal for the Maliau scenario is to model the ecosystem (i.e., intact primary..), which is slightly different than trying to validate the exact/perfect Maliau scenario.

arne-exe: One way to handle this is to use Maliau specific input data whenever possible, and use a range (if possible) of validation data for similar systems. During validation, this would then allow us to say where our Maliau simulation falls within the observed range of studies. Note that such a range of validation datasets will not be available for all output variables of interest ofcourse.

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+
+That said, if we validate soil on external dataset A and animal on external dataset B, and find that soil validates well but animal poorly, what does that mean? We cannot really conclude that soil is doing better than animal, because who knows if animal actually validates well on dataset A? Should we then all validate on the same dataset, which has a near-zero probability of having sampled all of our state variables? If it exists, we would have used it for input data already...
+[Nick has some thoughts on using global data for validation, i.e., if we validate the Maliau scenario onto global test dataset, and then redo for the SAFE scenario, then what does it mean to use the same global dataset twice?]{.aside}
```
arne-exe: This is challenging indeed. Perhaps it helps to separate this issue depending on whether you're trying to validate a model-specific process, or the overall VE simulation as a whole.
For model-specific processes, I think it's fine to use different datasets (i.e., your example of datasets A and B), because the goal here (I think) should be to calibrate the process/science within the specific model, we don't necessarily need to focus here on how the interaction with other models affects this process.
For validating the VE simulation as a whole, these interactions become more important, and the "goodness of fit" with the validation data will naturally vary depending on which validation datasets are used. I wonder if we can explore a way to capture this variation. Just brainstorming here, but perhaps we can quantify how this fit varies across outputs depending on how validation datasets change. Perhaps there is then a way to average the fit across outputs to get an overall idea of which validation datasets result in the best overall fit across all the outputs. Still a vague idea but writing it down in case anyone has some follow up ideas.

arne-exe: This could be visualized using a figure similar to the one below. Each color would be the result of changing 1 validation dataset (i.e., so all the others stay the same). Each option is then an outcome variable, so the figure would look a lot more complex, or we could focus on a few key output variables to keep it concise. We could organize the output variables in sections, so e.g. plant ones are located next to each other, colors of validation dataset changes can be more similar for datasets related to a specific model (e.g. plants). Visually, this would then show how changing e.g., a plant validation dataset impacts the validation for e.g. animal output variables.

Image

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+
+## Validate within or across modules?
+
+Currently we are validating within modules, but we are also increasingly aware of the need do so across modules. I'm leaning towards the latter because:
+
+1. VE's strength lies in its holistic nature. Validating within modules sort of defeats this original purpose.
```
arne-exe, line 101: Yes and no. Validation within models is necessary to guarantee the overall science within each model (without this you would assume that their theoretical frameworks are complete and without flaws, it may be but we should test this). There are likely quite a few shortcuts that we're taking due to missing things in the VE (because no time, too complicated, etc.). Validation within models gives us some control/oversight on this. I do get what you mean, if the only focus is validation within models, then we're unlikely to get the interactions right (unless we're lucky and somehow managed to nail the interactions perfectly on the first try 😄).

jacobcook1995: It's pretty hard to validate within a single model though, e.g. at the moment I can run the simulation and look at whether the soil results look plausible. You can pick out clearly incorrect behaviour (e.g. negative carbon pools), but "validating" the outputs really isn't possible because the subcanopy plants currently grow to form an impenetrable layer that block all light causing the soil to freeze

It is easier with some models (e.g. hydrology) because they are less linked to other models, but the biotic models are pretty tightly coupled

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+Do we then revisit `soil_c_pool_pom` only to get it right? Probably not, because (1) it also depends on `soil_c_pool_maom` and (2) the input data source probably contains both variables anyways. Most often than not we care about the whole carbon pool being partitioned correctly, rather than each pool hitting their own targets (if it happens, doesn't it just mean the former). In other words, since we will often be looking at groups of variables, then a validation target will often correspond to a group of variables. This naturally led us to think more about qualitative or emergent properties, e.g., that the ranking between soil nutrient pools should be preserved, soil fungi and bacteria should coexist, soil--microbe--plant stoichiometry should be preserved etc. And as you see it also naturally led us to think across modules.
+[There are other higher-level emergent properties that are not explicitly captured as state variables in VE, such as thinning rules, trophic distributions, network properties, energy compartments, biodiversity--productivity relationships etc., but they go beyond the discussion here.]{.aside}
```
jacobcook1995, line 131: I think the most relevant for soil is the carbon sequestration trajectories (i.e. is carbon being lost or stored by the soil over the time period in the manner the literature would lead you to expect).

I guess temporal properties like that are probably important for the plants as well (and I guess spatio-temporal ones for the animals)

analysis/soil/validation/conceptual_blocks.qmd
arne-exe: Writing down some more comments, they might not all make sense:

Discuss separating within model and whole system validation
For the whole system, the interactions are most important
Explore how variation in input data of specific models affects the interactions between models

Need to find a way to evaluate the interactions

Work in stages:
- first evaluate specific model science / functioning
- once we have shown that this is working well, we can focus on the interactions
- e.g. an increase in animal density may result in a lower biomass
- how can we use biomass validation datasets to determine the best fit in terms of animal density
- repeat this for each interaction between models in the VE
- then create a "final" interaction version, where the value of the interaction's driver reflects the most likely (i.e., "mean") value for the interaction's outcome
- for example, if we have 10 biomass validation datasets, which value for animal density (using VE) results in the mean biomass prediction across the 10 datasets?
- this would result in a set of input data that is calibrated in a way so that the interactions match what's most commonly observed for this type of ecosystem

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+[Nick has some thoughts on using global data for validation, i.e., if we validate the Maliau scenario onto global test dataset, and then redo for the SAFE scenario, then what does it mean to use the same global dataset twice?]{.aside}
+
+## Validate within or across modules?
+
+Currently we are validating within modules, but we are also increasingly aware of the need do so across modules. I'm leaning towards the latter because:
+
+1. VE's strength lies in its holistic nature. Validating within modules sort of defeats this original purpose.
+2. Validation within module overlooks the possibility that, for example, a poorly validated plant variable could be due to a poorly initialised or calibrated soil variable.
```
arne-exe, line 102: Working with parameter bounds or expected/realistic ranges may reduce the odds of this happening.
Parameters with unknown bounds may be more prone to this.

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+3. Validating across modules helps modules with hard-to-define targets. For example, plant validation datasets are relatively available but not animals, so why don't we just let animal use plant as validation targets? This is meaningful because if animal processes such as herbivory works well then we would expect the plant validation targets to do well too.
```
arne-exe, line 103: This can work in some cases, but not all. The risk here is that animal validation may "pass" when using plant validation datasets, without knowing why (think of it like a black box). This type of validation will likely leave gaps / unvalidated processes in the animal model (to use your example/analogy).

arne-exe: Not trying to criticize btw, but trying to help point out our potential blind spots.

analysis/soil/validation/conceptual_blocks.qmd
arne-exe, line 131: Good points, I agree that not every variable may be calibrated perfectly (or even that this is required to get "most" of the system's functioning right). What comes to mind here is to find a way to show which areas are calibrated well, and which ones are less (a bit like providing uncertainty estimates to a mean value). 
What we do need to be cautious off is to avoid giving people a chance to invalidate our "emergent conclusions" because we haven't sufficiently shown/proven the within model functioning (see my other comments above).

analysis/soil/validation/conceptual_blocks.qmd
arne-exe: > this would result in a set of input data that is calibrated in a way so that the interactions match what's most commonly observed for this type of ecosystem
>
This last one might be problematic, because the "ideal" input data was determined when only that specific input was changed, whereas what the comment above suggests is then to merge all these "ideal" input variables into one set. Because all of these "ideal" inputs will then interact with each other, it is uncertain how "ideal" they still are (i.e., how close to the mean observed outputs). I guess we could compare the difference between these new outputs and the "mean" ones, but not sure what exactly that would tell us. Need to think about it a bit more.

jacobcook1995 (approved):
These notes look sensible to me, only thing I had to add that was that soil dynamics probably genuinely are out of equilibrium, so whatever approach we decide on will probably need to focus more on trajectory rather than hitting specific values

analysis/soil/validation/conceptual_blocks.qmd
lelavathy, line 61: At this moment, there is no long-term station dataset available from Maliau specifically for validating the abiotic/hydrology modules. Because of that, I was thinking that using datasets from VJR and Danum could be a reasonable alternative, since those sites are also relatively intact tropical forests and environmentally quite similar to Maliau.

This would also align with the idea that the Maliau scenario is intended to represent a broader intact Bornean rainforest system rather than an exact replication of Maliau itself. Using validation datasets from comparable forest systems may therefore provide more robust insights than relying only on limited Maliau-specific observations.

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+[The Maliau scenario is monthly from 2010-01-01 to 2021-01-01.]{.aside}
+
+As a counter example, the soil input data had to be sourced from different study periods. And we do not have soil validation dataset collected exactly on the final date of the simulation. Hence the question in this section's heading.
+
+Even if abiotic inputs and outputs have a clear time mapping, the abiotic validation targets are affected by variables from other modules that do not (e.g., microclimate depends on plants). Because there is no spin-up phase in VE (except animals maybe), during the early time steps of simulation the state variables are likely doing *both* endogenic/internal stabilisation and reacting to exogenous factors. As such, even if we have clear validation targets, it is still hard to disentangle the effect of poor initialisation vs. poor parameter calibration.
```
lelavathy, line 33: Another related challenge may be the difference in response timescales across modules. Abiotic variables such as temperature, hydrology, or microclimate can respond relatively quickly to short-term climate forcing, whereas variables such as soil carbon pools, nutrient dynamics, or vegetation structure may evolve much more slowly over time. Because of this, validation/calibration approaches that are appropriate for fast abiotic processes may not necessarily work well for slower ecological processes, especially in the absence of a dedicated spin-up or equilibration phase.

jacobcook1995: 100%, this is basically why doing some kind of spin-up is reasonable for hydrology and microclimate, but not for soil where it can take 100s of years to reach a stable density of soil carbon (which is a real ecological effect, rather than a model artefact)

analysis/soil/validation/conceptual_blocks.qmd
lelavathy: I agree with both points here. Initially, I thought the abiotic/hydrology modules could be calibrated and validated more independently because they are more directly linked to physical processes and historical climate data. However, once the modules are coupled together, it becomes much harder to tell where unrealistic behaviour is actually coming from.

For example, hydrology or microclimate may behave reasonably on their own, but once vegetation and soil interactions are included, the outputs may change significantly. So I agree that within-module validation is still useful for identifying obviously unrealistic behaviour, but it may not be enough on its own to evaluate the behaviour of the fully coupled VE system.

analysis/soil/validation/conceptual_blocks.qmd
nickwctan, line 61: Yes I think the concensus of Maliau as an intact system is there and would serve as a validation reference. However, if we then validate Danum, a similar intact tropical forest system, how much would our validation datasets differ? I guess not so much.

analysis/soil/validation/conceptual_blocks.qmd
nickwctan: I think it would be across module for animals.

analysis/soil/validation/conceptual_blocks.qmd
```diff
@@ -0,0 +1,57 @@
...
+
+# Preamble
+
+- I learned about "validation" from the statistical modelling perspective; there may be some nuanced difference from process-based or mechanistic modelling.
+- I try to move across modules but apologies in advance if I have gotten anything wrong.
+- I try to think about validation generally, instead of only for Maliau itself. Solving this for Maliau = solving for any subsequent scenarios.
+
+# Conceptual blocks
+
+## How long should we run the simulation?
```
nickwctan, line 22: This is something I have to think about, do we validate our animal population across time, or just validate a snapshot of the population. Can we find empirical datasets of population or biomass fluctuations in function of time?

analysis/soil/validation/conceptual_blocks.qmd
nickwctan, line 131: At the moment for animals, I would say the collection of validation datasets and its ideas have been on emergent properties. They could be population turnover, growth rate, time to maturity, movement (which might be ambitious), trophic mass flow... I am currently looking at building a trophic network to potentially explore network properties and energy flow.

nickwctan (approved):

analysis/soil/validation/conceptual_blocks.qmd
Baizurah, line 22: From animal perspective, it will depends on the validation question and should not be decided as single number yet.  For instance, short runs are useful for functionality testing, as we can isolate whether the feeding, growth, mortality, predation does make sense. However, in certain cases, longer runs will be required as early timesteps may reflect internal adjustment of the animal rather than true ecological response. For site-specific scenario,  this is tricky because animal validation data are unlikely to match the exact simulation start and end dates, so we may need to focus more on emergent patterns over time rather than one final timestep.

analysis/soil/validation/conceptual_blocks.qmd
Baizurah: Thanks to you I Googled "spin-up phase", and this raises an important practical question related to current functionality testing that I've conducted recently. If the animal module naturally shows unstable early dynamics (e.g. predator decline or changing trophic interactions), how do we determine whether these patterns reflect expected stabilisation behaviour versus something that may need further investigation from an ecological or modelling perspective? I’m also wondering whether there are criteria or indicators we should use to identify when the system is sufficiently stabilised for interpretation or validation.

analysis/soil/validation/conceptual_blocks.qmd
Baizurah, line 61: My take on this is that Maliau datasets are definitely desirable and should remain the main reference for validation. But at the same time, as everyone already mentioned, expecting a strictly Maliau-only validation dataset is probably unrealistic. So I think using datasets from ecologically comparable systems is both reasonable and likely necessary where important gaps exist.

That said, from an ecological perspective, two “similar primary/intact forests” are never truly the same, especially from biodiversity perspective. Even if both are considered intact tropical forests, differences in animal behaviour, trophic structure, and FG dynamics can still presents due to landscape reserve size, surrounding landscape pressure, connectivity, hunting pressure, edge effects, disturbance history, sampling methods, and many other factors. Some of these finer-scale differences may not even be fully captured in the current setup.

So for the animal module, I think the important question is not simply whether another forest is “Maliau-like”, but whether it is similar enough for the specific validation target we are trying to assess.

For example, if we are validating broad trophic structure or general FG-level patterns, then forests such as Danum may be reasonable comparisons because both systems are broadly intact tropical forests. But if we move towards more sensitive outputs like predator persistence, herbivore biomass, movement, or predator-prey balance, then differences in reserve size, connectivity, hunting pressure, surrounding landscape, productivity, and disturbance history may become much more important.

So perhaps the first step is defining the validation target itself, then identifying which ecological factors are most likely to influence that target, and only after that deciding whether an external dataset such as Danum is appropriate.

analysis/soil/validation/conceptual_blocks.qmd
Baizurah: I think this works both ways. Within-module validation is still important to check whether specific processes are functioning sensibly, while across-module validation is needed to understand how those processes interact within the wider ecosystem

analysis/soil/validation/conceptual_blocks.qmd
Baizurah: I understand the concerns here, and like @arne-exe said, this could potentially work in certain cases, but it may also create confusion because a “good” plant result does not automatically mean the animal processes underneath are correct. For example, herbivore density might be too high, but feeding rate too low, and the plant outputs could still appear reasonable overall. so using plant validation may help to assess the interactions, but we still need to be sure, whether the parameter/constants used for such animal processes are realistic/not.

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Thanks. I noted down your point on "animal validation data are unlikely to match the exact simulation start and end dates".

You mentioned functionality testing, and I agree of the context. However, we might want to separate validation for testing vs. validation for prediction/forecasting. I feel that the latter is what we're aiming in an eventual paper, whereas the former is mostly for internal stuff to polish VE.

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Added snippets of these great feedback into the doc.

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Thanks, I have highlighted that we are probably treating "Maliau" as a representation of intact primary forest. I guess this is related to @annarallings ' question re: Silwood.

hrlai: Thanks I have added this alongside @arne-exe 's comment on the same topic.

hrlai: Thanks, I have summarised your point on forest idiosyncrasy and first steps into the doc.

hrlai: Same thought here, have added to the doc.

arne-exe: Very good points Bai, I agree that there might be two types of processes (i.e., those that can be seen as more general/similar across intact forests, and those that might show very distinct/local adaptations that may not be observed in similar intact forests). It's really hard to accurately represent the second set of processes in a general way that captures everything across "all" intact forests. 

There's a very important aspect here, which is how the VE handles fine-scale adaptations, for example at species level. Currently, the VE does not take species level adaptations into account as we're working with functional types/groups. This has benefits but also some limitations (as we're ignoring a lot of the intricate ways that species may cope with certain aspects). We've sort of already accepted this for modelling purposes, so perhaps we also need to "make a switch" in our head about what the VE can (and cannot) tell us about these fine-scale adaptations to specific processes. Here, it's not so much about the VE being 100% right, it's more about being clear about the limitations and uncertainty on the predictions that we can make.

To put this differently: let's say there is hypothetical process X (e.g. related to how plant species can access soil nutrients). Some species in Danum could show very distinct adaptations to this process, giving them a clear advantage over the ones in Maliau. Now, if we're calibrating/validating the VE using the data of Maliau, then it might be correct for Maliau, but not for Danum, even though they are both considered intact forests. To account for this, you'd have to get an understanding of which processes in your specific ecosystem show variability across spatial scales (i.e., variability may exist for tropical ecosystems but not for temperate systems). It's not even just spatial scales to consider, there are still a lot of unanswered science-related questions on the drivers of specific processes and how they may interact.

One way to deal with the above, once those processes have been determined (i.e., through existing studies and literature), is to account for this fine-scale adaptations within the code (i.e. by adding in specific "corrections" for a specific process (e.g. access to soil nutrients) depending on the ecosystem focused on (e.g. in the tropics). The specific corrections mentioned above could then be toggled on/off depending on what your ecosystem of focus is. The correction could then be triggered if, for example, the soil conditions of Danum are met (as then the specific adaptations are expected).  

Currently, the code/processes in the VE are assumed to work/capture most variations across all ecosystems. Because of this, it is quite natural that the outcomes of a simulation are unlikely to be a perfect match with validation datasets (unless everything is parameterized and validated using the same site, which again is not the main focus of the VE - the goal is not to provide a Maliau model, it is to provide a model that works globally for all types of ecosystem).

Without these corrections, we'd be comparing input data for different sites without accounting for these potential discrete adaptations. This is what we're doing now. The alternative is to not have any of such corrections, and to just accept that for some processes we are ignoring the fine-scale adaptations. This is hard pill to swallow from a science perspective, and ignoring these adaptations will naturally contribute to the differences between simulation outputs and validation datasets, but depending on the scope of the VE it may be the route we might need to take. We should discuss this with Rob, as it's a though decision to make, but at the moment the question is "just out there".

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Thanks for the notes. I think we should definitely revisit them at a later stage! I have added some of them into the doc.

The sensitivity analyses that @lelavathy and I are conducting will address some of your points above. In brief, there are local and global sensitivity analyses. The local approach ignores interactions, whereas the global approach accounts for interactions. The downside of a global sensitivity analysis (e.g., Sobol) is computational cost, but I am tackling that in PR [#254](https://github.com/ImperialCollegeLondon/ve_data_science/issues/254) and the HPC is helping tremendously. Validation is not sensitivity analysis, but what we learn from the latter should be useful for validation later.

For the visualisation, I have a good idea of how to do that as a multilayer network flattened to a pairwise matrix of within- and between-module metrics, and have discussed this with @annarallings and @jacobcook1995 recently. We could discuss this as a group in person soon.

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Thank you all for the great exchange. I have summarised them into the doc. 

My personal opinion is that validating within module does not guarantee the science within each model because the science of each model depends on other processes. As an example, even for a barebone resource--consumer food chain / food web model, the consumer dynamics still depends on the resource processes (e.g., basal plant growth), whether they are implicitly or explicitly modelled. When we isolate a module, the model becomes more and more phenomenological and hence departs further from the process-based design of VE.

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Yes, though setting parameter bounds is still a cross-module effort.

analysis/soil/validation/conceptual_blocks.qmd
hrlai: Thanks, I agree and have added these to the risk of sharing validation targets in the doc. I will also bear this in mind during the sensitivity analysis, because it is related to the interactions between variables (e.g., the cancelling-out effect that @Baizurah alluded to) in a global analysis.

hrlai (commented):
Thank you all for the feedback. I tried my best to summarise them into the doc. I am sure we will have a deeper discussion during the meeting.

arne-exe:
> Thank you all for the feedback. I tried my best to summarise them into the doc. I am sure we will have a deeper discussion during the meeting.

Thanks for initiating this and for summarizing the notes, very usefull!

Issue #165, opened 2026-01-20
Streamline (input) data file names for scenario runs


arne-exe:
As discussed with the data team, we should find a way to streamline file names so that they are easy to understand across users.
File names likely need to be unique for each scenario, although meta data (like grid and runtime) can be specified in either the sites folder or the assumptions folder (see #163). As an example, for Maliau we could potentially use something related to "Maliau scenario first run".

davidorme:
I'd use named scenarios that have some meaning - it makes it much easier to review the folders. That way you can also use numbers to show repeats of runs, if stochastic differences between runs become important. E.g.

`Maliau_single_pft_run_1`

You can also put the date on, but at least initially we're going to be producing wonky results that need a model fix. We wouldn't want to keep those results - we'd want to repeat it with the fixed code. So, the date isn't really meaningful.

We haven't started enforcing the idea of per folder manifest files, but those are intended to provide a way to log all the directories and provide information about them.

arne-exe:
> Ah, I see, but I don't store them in `data/derived` for soil and litter, because they are already in `data/scenarios`... Are you guys all doing it this way?
>
Originally that was what was agreed upon. All derived data from analyses was supposed to go into derived. 
The scenario folders are a new/recent addition to facilitate organizing and running the actual scenario.
All plant input data files are generated first to derived data, and then put in the Maliau_1 scenario folder.
I agree there is duplication, due to the tradeoff between having duplication and making it easier to run a scenario as a team that is working on different components (i.e., no single person is constructing the input data for all models by themself).

davidorme:
There's a lot in the responses - thanks for all the discussion! I'm going to try and pick up on a couple of points and see if I can align what I was thinking with your practical experience.

Again - I am _not_ at the sharp end of assembling these data, so any feedback really welcome. What I'm trying to do here is expand what I described above and try and set out how I think possible approaches fit into the likely usage patterns, particularly as we move to the HPC.

> Does it really? For testing scenarios I would just load the already existing data (or generate slightly altered versions using the analysis scripts) from the derived data folder. For the few "main" scenarios, like Maliau_1, Maliau_2,.. Silwood_1, Silwood_2,.., Panama_1, Panama_2,.. etc, I think having duplicated files in their respective folder is fine.

[...]

PR #321, opened 2026-05-11
Python-based VE TOML configuration generation workflow

lelavathy:
## Summary

This PR adds a Python-based workflow for generating modular configuration files for Maliau simulation scenarios.

This PR is related to and extends ideas introduced in:
PR #289 (tool to generate ve run config files)

The implementation adapts similar concepts into Python-based generation of modular configuration files

### Added

* `build_modular_config_maliau.py`

  * Interactive VE scenario selection
  * Generation of modular TOML configuration files
  * Dataset variable insertion from data_variables.py
  * Plant constants loading from CSV FILE

* Helper functions in `tools/python/build_config.py`

  * Scenario loading
  * Scenario display and selection
  * Grid suffix generation
  * TOML writing utilities
  * Data variable block generation

* Variable definition module:

  * `tools/python/data_variables.py`

### Purpose

This workflow improves maintainability and reduces repetitive manual TOML editing for generating modular configuration files for VE simulations.

### Generated configuration files

* ve_run.toml
* data_config.toml
* animal_config.toml
* plants_config.toml
* soil_microbial_groups.toml

### Notes
-The workflow currently targets Maliau Basin VE scenarios.
-Dataset file paths are currently placeholders and may require updating depending on local workflow structure.
-Variable definitions are imported dynamically from helper modules to minimise repetitive configuration code.

### DIFF
--- a/analysis/site/build_modular_config_maliau.py
+++ b/analysis/site/build_modular_config_maliau.py
@@ -19,7 +19,7 @@ description: |
     - soil_microbial_groups.toml
 
   The script also:
-    - loads plant constants from CSV files
+    - loads plant constants from a CSV file
     - generates scenario-specific file naming
     - creates modular VE configuration structures
     - automatically writes TOML files to scenario output directories
--- a/analysis/site/build_modular_config_maliau.py
+++ b/analysis/site/build_modular_config_maliau.py
@@ -1,0 +1,622 @@
+"""
+---
+title: Build VE Modular Configuration Files for Maliau Scenarios
+
+description: |
+  This script generates Virtual Ecosystem (VE) modular configuration
+  TOML files for predefined Maliau Basin simulation scenarios.
+
+  The workflow loads scenario-based grid and timing configurations
+  from a TOML file, allows users to interactively select a scenario,
+  and automatically builds the required VE configuration files
+  needed to run ecosystem simulations.
+
+  Generated configuration files include:
+    - ve_run.toml
+    - data_config.toml
+    - animal_config.toml
+    - plants_config.toml
+    - soil_microbial_groups.toml
+
+  The script also:
+    - loads plant constants from CSV files
+    - generates scenario-specific file naming
+    - creates modular VE configuration structures
+    - automatically writes TOML files to scenario output directories
+
+  Variable definitions for climate, soil, litter, and plant
+  datasets are imported from helper utility modules to improve
+  maintainability and reduce repetitive TOML-writing code.
+
+virtual_ecosystem_module:
+  - Abiotic
+  - Hydrology
+  - Plant
+  - Soil
+  - Animal
+
+author:
+  - Lelavathy Samikan
+
+status: final
+
+input_files:
+  - name: maliau_grid_definition.toml
+    path: data/derived/site/maliau/maliau_grid_definition.toml
+    description: |
+      Scenario definition file containing VE grid structure,
+      spatial configuration, and simulation timing settings.
+
+  - name: plant_constants_Maliau_<grid_suffix>.csv
+    path: data/data/plant_constants_Maliau_<grid_suffix>.csv
+    description: |
+      CSV file containing plant model constants used
+      for VE plant module configuration.
+
+output_files:
+  - name: ve_run.toml
+    path: data/scenarios/maliau/<scenario>_config/
+    description: |
+      Main VE run configuration file specifying
+      enabled ecosystem modules.
+
+  - name: data_config.toml
+    path: data/scenarios/maliau/<scenario>_config/
+    description: |
+      VE data configuration file containing climate,
+      elevation, soil, and litter variable definitions.
+
+  - name: animal_config.toml
+    path: data/scenarios/maliau/<scenario>_config/
+    description: |
+      Configuration file for VE animal module settings
+      and cohort export options.
+
+  - name: plants_config.toml
+    path: data/scenarios/maliau/<scenario>_config/
+    description: |
+      Configuration file for VE plant module settings,
+      plant input datasets, and plant constants.
+
+  - name: soil_microbial_groups.toml
+    path: data/scenarios/maliau/<scenario>_config/
+    description: |
+      Configuration file defining soil microbial
+      functional groups and enzyme classes.
+
+package_dependencies:
+  - pathlib
+  - pandas
+  - toml
+
+usage_notes: |
+  - Ensure that scenario definitions are properly specified in
+    `maliau_grid_definition.toml` under the scenario section
+    (e.g., `maliau_1`, `maliau_2`).
+  - Ensure that all required NetCDF input datasets and CSV parameter
+    files exist in data folder before running the script.
+  - Update placeholder file paths for climate, elevation, soil,
+    litter, and plant datasets if necessary.
+  - In the terminal, run this script from the project root directory
+    using:
+
+      python -m analysis.site.build_modular_config_maliau
+
+  - Select a scenario interactively by typing its name
+    (e.g., `maliau_1`, `maliau_2`) or the corresponding number
+    when prompted.
+  - The script will load the selected scenario configuration and
+    automatically generate the required VE TOML configuration files.
+  - Output configuration files will be written to:
+
+      data/scenarios/maliau/<scenario>_build_config/
+
+  - To generate configuration files for another scenario,
+    rerun the script and select a different scenario.
+
+  Run as: python -m analysis.site.build_modular_config_maliau
+---
+"""  # noqa: D400, D212, D205, D415
+
+
+# =============================================================================
+# Import packages
+# =============================================================================
+# Import required Python libraries for:
+#   - file handling
+#   - TOML parsing
+#   - CSV processing
+#   - VE helper utilities for configuration buildings.
+
+from pathlib import Path
+
+import pandas as pd
+
+from tools.python.build_config import (
+    add_data_variables,
+    get_grid_suffix,
+    load_scenario,
+    select_scenario,
+    write_toml,
+)
+from tools.python.data_variables import (
+    climate_variables,
+    litter_variables,
+    plant_variables,
+    soil_variables,
+)
+
+# =============================================================================
+# Scenario definition file
+# =============================================================================
+# Define the scenario-based VE configuration TOML file
+# containing multiple Maliau simulation scenarios.
+
+scenario_file = "data/derived/site/maliau/maliau_grid_definition.toml"
+
+# =============================================================================
+# Select VE scenario
+# =============================================================================
+# Display available VE scenarios and allow
+# the user to select one interactively.
+
+scenario = select_scenario(scenario_file)
+
+# =============================================================================
+# Load selected scenario
+# =============================================================================
+# Load VE grid and timing configuration
+# from the selected scenario.
+
+scenario_core = load_scenario(
+    scenario_file,
+    scenario,
+)
+
+core_grid = scenario_core["grid"]
+
+core_timing = scenario_core["timing"]
+
+# =============================================================================
+# Add canopy structure settings
+# =============================================================================
+# Define canopy structure parameters required
+# for VE vegetation and abiotic simulations.
+
+core_grid["canopy_layers"] = 10
+
+core_grid["above_canopy_height_offset"] = 2.0
+
+core_grid["subcanopy_layer_height"] = 1.5
+
+core_grid["surface_layer_height"] = 0.1
+
+# =============================================================================
+# Generate grid suffix
+# =============================================================================
+# Create scenario-specific grid suffix used
+# for naming VE datasets consistently.
+
+grid_suffix = get_grid_suffix(core_grid)
+
+print(f"\nselected scenario : {scenario}")
+print(f"grid size         : {grid_suffix}")
+
+# =============================================================================
+# Output directory (Manually edit if necessary)
+# =============================================================================
+# Create scenario-specific output directory
+# for generated VE configuration files.
+
+output_dir = Path(f"data/scenarios/maliau/{scenario}_config")
+
+output_dir.mkdir(
+    parents=True,
+    exist_ok=True,
+)
+
+# =============================================================================
+# Dataset file paths (Manually edit if necessary)
+# =============================================================================
+# Define VE dataset paths for:
+#   - climate
+#   - elevation
+#   - soil
+#   - litter
+#   - plant inputs
+
+climate_file = f"../data/era5_maliau_{grid_suffix}_2010_2020.nc"
+
+elevation_file = f"../data/elevation_maliau_{grid_suffix}.nc"
+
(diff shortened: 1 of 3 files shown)


analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+
+  Generated configuration files include:
+    - ve_run.toml
+    - data_config.toml
+    - animal_config.toml
+    - plants_config.toml
+    - soil_microbial_groups.toml
+
+  The script also:
+    - loads plant constants from CSV files
```
hrlai:
```suggestion
    - loads plant constants from a CSV file
```

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+  Variable definitions for climate, soil, litter, and plant
+  datasets are imported from helper utility modules to improve
+  maintainability and reduce repetitive TOML-writing code.
+
+virtual_ecosystem_module:
+  - Abiotic
+  - Hydrology
+  - Plant
+  - Soil
+  - Animal
```
hrlai, line 36: I have been writing "All" simply if I am refering to all modules, this saves you from having to type "abiotic_simple" or even "core" which are strictly also modules...

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+
+# =============================================================================
+# Load selected scenario
+# =============================================================================
+# Load VE grid and timing configuration
+# from the selected scenario.
+
+scenario_core = load_scenario(
+    scenario_file,
+    scenario,
```
hrlai, line 173: So it seems that it is not necessary for the user to interactively select `scenario`, right? I could also just assign `scenario = "maliau_1"` if I know it? If so, then maybe add a short note that using `select_scenario()` is optional?

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+# for VE vegetation and abiotic simulations.
+
+core_grid["canopy_layers"] = 10
+
+core_grid["above_canopy_height_offset"] = 2.0
+
+core_grid["subcanopy_layer_height"] = 1.5
+
+core_grid["surface_layer_height"] = 0.1
+
```
hrlai, line 193: These seems to be the default values that will be set by VE if the user does not supply anything. I understand that you are going for a full(er) config build, but what happens if I do not supply these defaults values in this script? Will the whole workflow still work?

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+# =============================================================================
+# Generate grid suffix
+# =============================================================================
+# Create scenario-specific grid suffix used
+# for naming VE datasets consistently.
+
+grid_suffix = get_grid_suffix(core_grid)
+
+print(f"\nselected scenario : {scenario}")
+print(f"grid size         : {grid_suffix}")
```
hrlai, line 203: This is a good place for the team @Baizurah @nickwctan @arne-exe @annarallings to discuss whether we treat "maliau_1_50x50" or "maliau_1" as a Scenario name.

To me, "maliau_1_50x50" as a whole should be the scenario name for two reasons:

1. "maliau_1" will only ever refer to "maliau_1_50x50", i.e. the two are synonymous
2. eventually we might have a temporal suffix for shorter vs. longer runs, so then you would have to maintain the code here to extract and print a timing suffix

There is nothing with the process here (and can feel a bit of a semantic), but I'm exploring if things could be simpler for you to maintain here.

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+with open(
+    data_config_file,
+    "w",
+) as file:
+    file.write(data_config_text)
+
+print(f"written: {data_config_file}")
+
+# =============================================================================
+# Generate animal _config.toml
```
hrlai:
```suggestion
# Generate animal_config.toml
```

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+# =============================================================================
+# Generate the VE animal module configuration file
+# including cohort export settings and attributes.
+
+animal_config_text = """[animal]
```
hrlai, line 355: in the lines above, you have a more flexible way of specifying config variables, but for the animal config they are harded coded in as a text chunk. Do you think this should mimic the above?

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+functional_group_definitions_path = '../data/animal_functional_groups_Maliau_level1.csv'
```
hrlai, line 356: For example, one might want to select a level 2 FG data for animals.

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+# =============================================================================
+# Build VE plant module configuration file
+# including plant datasets and constants.
+
+plants_config_text = f"""[plants]
+cohort_data_path = "../data/plant_cohort_data_Maliau_{grid_suffix}.csv"
+pft_definitions_path = "../data/plant_pft_definitions_Maliau_{grid_suffix}.csv"
+
+[plants.community_data_export]
+required_data = ["cohorts","community_canopy","stem_canopy"]
```
hrlai, line 412: Like above, I think things like `plants.community_data_export.required_data` could also be soft coded so the user can change the selection.

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+
+print(f"written: {plants_config_file}")
+
+# =============================================================================
+# Generate soil_microbial_groups.toml
+# =============================================================================
+# Generate the VE soil microbial functional group
+# and enzyme class configuration definitions.
+
+soil_microbial_groups_text = """[[soil.microbial_group_definition]]"
```
hrlai, line 453: This is a similar situation to that of animal above, but let's leave it as a TOML for now. @jacobcook1995 is seeing how we could switch csv files like the plant and animal input data.

tools/python/build_config.py
```diff
@@ -0,0 +1,373 @@
...
+              - grid configuration
+              - timing configuration
+
+    Raises:
+        ValueError:
+            If the requested scenario is not found.
+
+    """
+
+    # Load TOML configuration file
```
hrlai, line 126: Replace this as "site definition" or "site core definition" to distinguish it from "VE configuration" files?

tools/python/build_config.py
```diff
@@ -0,0 +1,373 @@
...
+        start=1,
+    ):
+        scenario_info = scenarios[scenario_name]
+
+        # Access VE grid configuration
+        grid = scenario_info["core"]["grid"]
+
+        cell_nx = grid["cell_nx"]
+
+        cell_ny = grid["cell_ny"]
```
hrlai, line 201: Just a thought, reading these lines make me think that we should always ensure that the grid suffix (or timing suffix in the future) of scenario names should be the same as `cell_nx_x_cell_ny` for some automated config building in the future. We are already doing this subconsciously.

tools/python/data_variables.py
```diff
@@ -0,0 +1,120 @@
...
+
+---
+"""  # noqa: D400, D212, D205, D415
+# =============================================================================
+# VE variable definitions
+# =============================================================================
+# Define grouped VE variable names used for
+# automatic TOML configuration generation.
+
+climate_variables = [
```
hrlai, line 70: Just a note: in the future I plan to automatically generate these list so they can be more easily maintained with changes on the VE side. It should be easier here because in Python you can simply import the list of required input variables from `virtual_ecosystem`. We could give the coding team a reminder to implement this? @jacobcook1995 is already aware of the value of this.

hrlai (commented):
Mostly looks good to me and I have just added some comments for discussion.

analysis/site/build_modular_config_maliau.py
nickwctan: I am fine with either :)

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+print(f"grid size         : {grid_suffix}")
+
+# =============================================================================
+# Output directory (Manually edit if necessary)
+# =============================================================================
+# Create scenario-specific output directory
+# for generated VE configuration files.
+
+output_dir = Path(f"data/scenarios/maliau/{scenario}_config")
+
```
nickwctan, line 212: #TODO: remind ourselves to change the path here if we follow David's folder structure for HPC #253

analysis/site/build_modular_config_maliau.py
```diff
@@ -0,0 +1,622 @@
...
+# Build VE core data configuration including:
+#   - grid settings
+#   - timing settings
+#   - canopy structure settings
+#   - climate variables
+#   - elevation variables
+#   - soil variables
+#   - litter variables
+
+data_config_text = f"""[core.grid]
```
nickwctan, line 282:
```suggestion
data_config_text = f"""
[core.grid]
```

Just a suggestion to make the code looks cleaner, but it does introduce a new (blank) line on the toml file. I am fine with any choice. 

A quick search found that we can use [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html) to keep Python indentation readable

analysis/site/build_modular_config_maliau.py
lelavathy: Thank @hrlai  — these comments were really helpful and gave me a lot to think about regarding the longer-term structure and maintainability of the workflow.

> I agree this is worth discussing as a team. Personally, I slightly prefer names like `maliau_50x50_2010_2020` because they immediately communicate the spatial and temporal setup being used.
>
> I’m still wondering whether `maliau_1` is actually functioning more like a *site definition* rather than a true *scenario* definition. The term “scenario” could be reserved for actual experimental or modelling cases. For example:
>
> * `baseline`
> * `climatechange_rcp85`
> * `elevated_co2`
> * `drought_experiment`
>
> In that structure, the site definition might describe:
>
> * `maliau_50x50_years`
> * `maliau_10x10_years`
> * `safe_50x50_years`
>
> while the scenario would define the simulation experiment itself.
>
> So I think there may eventually be value in separating:
>
> * **site/grid definition**
> * **simulation scenario**
> * possibly also **runtime duration/configuration**
>
> even though currently they are somewhat blended together in the workflow.

This might also relate to issue #165 (Streamline (input) data file names for scenario runs), since clearer separation between site definitions and simulation scenarios could help make naming conventions and automated file generation more consistent in the future.

analysis/site/build_modular_config_maliau.py
lelavathy: I think this could become clearer once the distinction between *site definitions* and *simulation scenarios* is finalised.

At the moment, `maliau_1` feels more like site/grid information, since it mainly contains spatial and temporal metadata that can be read directly from `maliau_grid_definition.toml`. In that case, `select_scenario()` may actually be functioning more as a site-selection helper that simply displays the available site/grid definitions.

If we later separate:

* site/grid definitions
* simulation scenarios
* 
then I could imagine:

* the site being selected automatically from the available entries in `maliau_grid_definition.toml`
* while the actual simulation scenario (e.g. `baseline`, `climatechange_rcp85`) could optionally be selected by the user separately.

I think the naming/selection logic may become clearer once the site vs scenario distinction is more formally defined.

analysis/site/build_modular_config_maliau.py
nickwctan: The way I see how Lela intended for this script to be used, is for the user to modify the codes directly in `build_modular_config_maliau.py`  in their local drive as they wish according to what they want?

nickwctan: The other alternative I can think of is to make this whole script as functions (like in build_config.py) in order for it be soft coded. So then the script would fall under tools/python.

nickwctan (commented):
Neat, thanks Lela. I've added my suggestions. 

I initially thought you were building a full_config generator instead of a modular one :D

analysis/site/build_modular_config_maliau.py
lelavathy: Yes, these currently match VE default values, so the workflow should still work even if they are omitted here because VE will populate them internally.

I originally included them mainly to keep the generated configuration files fully explicit and reproducible, especially for debugging where having all parameter values written directly into the generated TOML files can make runs easier to trace and compare later.

At the moment, the script is using:

```python
core_grid["canopy_layers"] = 10
```

which always overwrites the value, even if the user already supplied a different value in the site definition TOML. For example, if a future site definition contains:

```toml
canopy_layers = 20
```

the current script would still replace it with `10`.

Using:

```python
core_grid.setdefault("canopy_layers", 10)
```

changes the behaviour slightly. It only inserts the value if the key does not already exist. So:

* if the site definition TOML does not specify `canopy_layers`, the script will add the default value (`10`)
* if the user already supplied a value in the TOML, that value will be preserved

I think this is probably a better compromise because it:

* preserves VE defaults if nothing is supplied
* avoids unintentionally overwriting user-defined settings
* makes the workflow more flexible for future site-specific canopy structures or scenario-specific vertical configurations

This could become more useful later if different sites or experiments require different canopy layer structures while still keeping the config-building workflow automated.

analysis/site/build_modular_config_maliau.py
nickwctan: Is there any difference in structure with your [R config builder maliau](https://github.com/ImperialCollegeLondon/ve_data_science/pull/289/changes#diff-fc3f9b27cc95873b9182f76cdf43910fdd9d76ff7306cc18ce2d0dbd85ebbec0)

analysis/site/build_modular_config_maliau.py
hrlai: Yeap, so I'm happy for you to keep it as it is here, what I wrote is more of a note for the future :)

analysis/site/build_modular_config_maliau.py
hrlai: In the R version, the nuance is that I specified the animal settings as a `list` object rather than a plain TOML text:

https://github.com/ImperialCollegeLondon/ve_data_science/blob/a458b146ab04a770b08b9ad04e66dce80f906862/analysis/site/build_config_maliau.R#L105-L136

The list is then converted to TOML by the `build_config` function. The idea is to let users collect their settings into a native R object, i.e., a list here, rather than having to write plain TOML text. The additional benefit of collecting settings into a list in R is so that you can refer any list item to any source objects, rather than having to hard code them into plain text. Sorry my explanation may be a bit convolution but I hope the gist is clear...

nickwctan: ah yes, the equivalent of this will be a dictionary in python.

Issue #276, opened 2026-04-16
Animal - Trophic interaction output


nickwctan:
When running simulations for 2 years in #274 with functional groups from the example data, it appears that all trophic interactions ceased at 4th time_index while the model kept running towards the end (24 timesteps).

At the same time, it seems like animals are decreasing in numbers at each timestep. The large body size FG die off at 4th time_index, while the smaller size FGs such as butterfly and caterpillar, survived to the end. 

Why is there a cease in the trophic interactions after the 4th time_index even when there are still butterfly and caterpillar FGs...?

Funny note: At the 30x30 grid run, the `trophic interaction output csv ` file was so huge that it exceeded the number of rows open office can handle 😆

PR #330, opened 2026-05-18
329 examine animal continuous state variables from `maliau_2`

### LINKED ISSUE #329: Examine animal continuous state variables from `maliau_2`

### LINKED ISSUE #501: Update tidy_continuous_data() to read Zarr outputs
`tidy_continuous_data` still reads netCDF outputs but VE has migrated to Zarr outputs. This is blocking PR #330

hrlai:
This PR aims to report and discuss the meaning of animal continuous state variables from the `maliau_2` scenario that hovered around zero. Understanding it will solve some roadblocks I'm having with the soil sensitivity analysis in #300 .

@jacobcook1995 we might need your expertise too, since the consumption rates are soil-related. 

Please see the rendered notebook [here](https://github.com/ImperialCollegeLondon/ve_data_science/blob/329-examine-animal-continuous-state-variables-from-maliau_2/analysis/animal/continuous_states/rendered/small_temporal_variation.md).

Thanks!


analysis/animal/continuous_states/small_temporal_variation.qmd
```diff
@@ -0,0 +1,133 @@
...
+I ran the full `maliau_2` scenario:
+
+- config in `data/scenarios/maliau/maliau_2/config`
+- data in `data/scenarios/maliau/maliau_2/data`
+- VE version: v0.1.1a17 (dev version; commit [b6a216](https://github.com/ImperialCollegeLondon/virtual_ecosystem/commit/b6a216bd263a50adfb5e1c90eb64583699578017))
+- OS: Windows 11
+
+::: {.callout-note}
+# Improvement
+I'm thinking of writing a function to automatically summarise key model configurations from the full config TOML compiled by `ve_run`. We know what `maliau_2` means but others outside our team may not remember. Sometimes a scenario config may also change. Tracked in [this Issue](https://github.com/ImperialCollegeLondon/ve_data_science/issues/327).
```
nickwctan: sounds like a good idea but doesn't it get quite lengthy?

hrlai: It can be, so we'll need to discuss what to report minimally, maybe only grid number, grid size, time length. Let's discuss this during a meeting one day.

analysis/animal/continuous_states/small_temporal_variation.qmd
```diff
@@ -0,0 +1,133 @@
...
+  "animal_saprotrophic_fungi_consumption",
+  "total_animal_respiration"
+)
+animal_cont <- tidy_continuous_data(
+  here("data/scenarios/maliau/maliau_2/out/all_continuous_data.nc"),
+  variables = animal_vars
+)
+```
+
+First I saw that the range of these state variables are very small. Are they truly very small, or are they numerical imprecisions that need to be zapped to zero?
```
nickwctan: Something I wanted to look back at for some time. I remember my previous conversation with @TaranRallings that these small numbers were expected. Maybe this is a question for Taran or Jacob. The symmetrical oscillation of consumption around zero through time is strange. 

Have you looked at how much biomass of these resource pools are, in comparison to the consumption? Sorry for bouncing back another question at you.

hrlai: > Have you looked at how much biomass of these resource pools are, in comparison to the consumption? Sorry for bouncing back another question at you.

No that is a good question because rates will be low if there was little resource to consume in the first place.

analysis/animal/continuous_states/small_temporal_variation.qmd
```diff
@@ -0,0 +1,133 @@
...
+  here("data/scenarios/maliau/maliau_2/out/animal_cohort_data.csv")
+)
+max_cohort_time <- max(animal_cohort$time_index)
+```
+
+Before proceeding, I checked the animal cohort data and saw that all cohorts went extinct after time step `r max_cohort_time`.
+
+A few follow-up questions upon seeing the temporal graphs:
+
+- Why do we still see non-zero values in some variables long after all animals have gone extinct since time step `r max_cohort_time`?
```
nickwctan: I imagine you are only talking about the variables you mentioned in this PR? Currently most things die off already at the 4th timestep cause there are no plant resource to animal pipeline yet.

analysis/animal/continuous_states/small_temporal_variation.qmd
```diff
@@ -0,0 +1,133 @@
...
+- Presumably these variables are positive only; what do the negative values mean? The way they fluctuate almost symmetrically around zero makes me suspect that the non-zero values are not true non-zeros but numerical imprecision.
```
nickwctan: Weird! Yes negative consumption is definitely not normal. However, can there be negative values for mycorrhizal fungi in the data object? If so, they are treated as zero for animal consumption purposes...

analysis/animal/continuous_states/small_temporal_variation.qmd
hrlai: @nickwctan  I have added the resource dynamics, looks like you're onto something! Keen to hear your thoughts.

analysis/animal/continuous_states/small_temporal_variation.qmd
hrlai: Yes, only the continuous state variables used in this PR, which also had animals died off at the 4th time step (during the meeting I mentioned the 3rd time step, but this has been fixed; I forgot that the time index starts from zero, so I added 1 to it).

So my original question is why do we have non-zero consumption rates when there is no animal after the 4th time step.

nickwctan (commented):
I had a quick look and left some comments. I feel like Taran or Jacob are probably the better person to answer the questions you have. Or maybe Bai has some insights on this?

analysis/animal/continuous_states/small_temporal_variation.qmd
nickwctan: > why do we have non-zero consumption rates when there is no animal after the 4th time step.

That is the question I wonder as well! :D The animal interaction output stops producing output after the 4th timestep too #276 

Which functional_group.csv file you are using for your animal input file?

analysis/animal/continuous_states/small_temporal_variation.ipynb
```diff
@@ -0,0 +1,331 @@
+{
```
jacobcook1995: <p>"At the end of this report, I explain why we might want to design a scenario".</p><p> </p><p>I guess the essential problem here is that the results of any parameter sensitivity are not going to remain valid until the model is fully calibrated (i.e. have reached a stable state in the code, using input data and parameter estimates that we at least somewhat trust)</p><p> </p><p>Isn't a huge issue if we are just looking to get a basic sense of what matters, and have written pipelines that mean analyses can be rerun with updated versions</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ImperialCollegeLondon/ve_data_science/pull/330/discussion/#comment-3265264837'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jacobcook1995: <p>Pretty certain (as some of these are below zero) that these are numerical imprecisions</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ImperialCollegeLondon/ve_data_science/pull/330/discussion/#comment-3265264841'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

jacobcook1995: <p>"Mainly so that we can include animal-related state variables into the sensitivity analyses." </p><p> </p><p>I think this will be a persistent problem with sensitivity analysis pre-calibration, because even if you have persistent animal populations, I strongly suspect that they will be funky, and even if they aren't there's a lot wrong in the plant model (then there's almost certainly uncaught bugs)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ImperialCollegeLondon/ve_data_science/pull/330/discussion/#comment-3265264850'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

Baizurah: <p>I agree with <span class="mention" data-value="jacobcook1995" data-denotation-char="@">﻿<span contenteditable="false">@jacobcook1995</span>﻿</span> . Commenting from here, just to experiment with different way of this can works. So amazing! </p>

Baizurah: <blockquote>Should we consider an alternative set of animal FG definitions? Currently <code>maliau_2</code> uses the level 1 definition, which contain only a single herbivorous endotherm that always go extinct very early on. Has anyone run VE with the level 2 definitions? If the level 2 groups also go extinct, should we consider an alternative set (perhaps more basal in tropic levels) that can persist over time, and hence continue to keep the animal and non-animal components coupled until the end of simulation?</blockquote><blockquote><br></blockquote><p>I did try using different input files apart from what we currently have in Globus, as part of a sense-check experiment where I controlled traits, including testing different body sizes of herbivores feeding on the same resources. In short, all runs still ended up with animals declining rapidly by around t=4, although it is worth noting that I ran them using the pip version 17. I’m not sure Level 2 definitions would necessarily lead to better outputs at this stage, because if the simpler FG setup (Level 1) does not work, I doubt Level 2 would automatically improve things, except perhaps making the outputs more difficult to interpret due to the higher number of FGs involved. I’m also curious what you mean by “more basal in trophic levels”? <span class="mention" data-value="hrlai" data-denotation-char="@">﻿<span contenteditable="false">@hrlai</span>﻿</span> </p>

Baizurah: > > Should we consider an alternative set of animal FG definitions? Currently `maliau_2` uses the level 1 definition, which contain only a single herbivorous endotherm that always go extinct very early on. Has anyone run VE with the level 2 definitions? If the level 2 groups also go extinct, should we consider an alternative set (perhaps more basal in tropic levels) that can persist over time, and hence continue to keep the animal and non-animal components coupled until the end of simulation?
> 
> > 
> 
> I did try using different input files apart from what we currently have in Globus, as part of a sense-check experiment where I controlled traits, including testing different body sizes of herbivores feeding on the same resources. In short, all runs still ended up with animals declining rapidly by around t=4, although it is worth noting that I ran them using the pip version 17. I’m not sure Level 2 definitions would necessarily lead to better outputs at this stage, because if the simpler FG setup (Level 1) does not work, I doubt Level 2 would automatically improve things, except perhaps making the outputs more difficult to interpret due to the higher number of FGs involved. I’m also curious what you mean by “more basal in trophic levels”? ﻿@hrlai

Wonder why commenting through "ReviewNB" is not visible here.... @hrlai

analysis/animal/continuous_states/small_temporal_variation.qmd
hrlai: The level 1 `data/scenarios/maliau/maliau_2/data/animal_functional_groups_Maliau_level1.csv`

This is by the way a good example of why we need to include some model run info in the report.

analysis/animal/continuous_states/small_temporal_variation.ipynb
hrlai: > I’m also curious what you mean by “more basal in trophic levels”?

I meant something more like a small-bodied herbivorous consumer, which does not require much resource to maintain a viable population.

> Wonder why commenting through "ReviewNB" is not visible here.... 

Not sure. From my end, there is no easy way to tell whether you commented via ReviewNB or directly here, because the comments will show on both sides....

analysis/animal/continuous_states/small_temporal_variation.qmd
nickwctan: > @nickwctan I have added the resource dynamics, looks like you're onto something! Keen to hear your thoughts.

Thanks! It seems to me that feeding activity corresponds to the resource availability temporally. However, it does not explain the very small values we have in biomass consumption, which makes me think that its a result of numerical imprecision. I'll have a chat with Taran about this.

analysis/animal/continuous_states/small_temporal_variation.ipynb
nickwctan: That begs the question of whether we should zap these numerical imprecisions into zero? Is it safe to assume that there is no consumption going on?

nickwctan:
> do these runs show similar results with the foraging intensity on plant resources? carcasses? pred-prey interactions?

Yes, to do that, you'd have to run the example animal functional group in the VE example. 

> [Nick (@nickwctan)](https://github.com/nickwctan) is this related to the visualisation tools that you're developing?

Yes #243 , its almost there, but feel free to send me the `trophic_interaction.csv` to me after your run or you can make a simple script to do the same thing

analysis/animal/continuous_states/small_temporal_variation.qmd
hrlai: The filename has been added to the latest notebook.

analysis/troubleshoot/animal_continuous_states/small_temporal_variation.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Why do we need persistent animal populations
+
+::: {.callout-caution collapse="false"}
+
+## Outdated
+
+Mainly so that we can include animal-related state variables into the sensitivity analyses. More importantly, the animal variables feed back into the non-animal variables. Unless we are truly aiming for an empty-forest scenario, we will be left with a half-complete sensitivity analysis.
+
+Should we consider an alternative set of animal FG definitions? Currently `maliau_2` uses the level 1 definition, which contain only a single herbivorous endotherm that always go extinct very early on. Has anyone run VE with the level 2 definitions? If the level 2 groups also go extinct, should we consider an alternative set (perhaps more basal in tropic levels) that can persist over time, and hence continue to keep the animal and non-animal components coupled until the end of simulation?
```
Baizurah, line 295: Check out #496, might be relevant to this. 
@hrlai

Baizurah:
Curious as to why `total_animal_respiration` remains zero exactly throughout, although we have feeding and cohorts. Veru useful report on model-behaviour pattern, perhaps if we are investigating this further, we can plot earthworms no of individuals+ population trends + resource consumption outputs to see if they die from feeding limitation or not. @hrlai

hrlai:
> Next thing to troubleshoot is on the FG persistence as it seems like earthworms are dying off even though resources are available.

Thanks @nickwctan , let's explore this later, I'm also curious what happened 😄

PR #339, opened 2026-05-22
Set up soil and litter validation screening and database

hrlai:
This PR sets up a workflow to screen, add and harmonise datasets for validation. **It is intended to work with the soil and litter modules only**, though it is open for adoption to other modules. This PR is expected to address ImperialCollegeLondon/virtual_ecosystem#334 ImperialCollegeLondon/virtual_ecosystem#336 ImperialCollegeLondon/virtual_ecosystem#337 ImperialCollegeLondon/virtual_ecosystem#338 

> [!TIP]
>  For background and rationale, it's easier to read the full documentation of my intended workflow [here](https://github.com/ImperialCollegeLondon/ve_data_science/blob/334-soil-and-litter-validation-search/docs/validation_database.md). The "Data screening workflow" section mentioned about an interactive table for screening, it is a report written in Qmd, you can see the compiled html table on [Dropbox](https://www.dropbox.com/scl/fi/lisq0gvxo1w9jd1ojpw3u/dataset_screening.html?rlkey=0osj23xrwv95fbgljv7qmrtqf&dl=0). **This HTML file is not intended to be pushed to GitHub, instead I needed an interactive table for my own workflow.**

Below is an overview of the changed files. I hope to merge this PR before it blows up further; I will split the follow-up data addition and harmonisation into a future PR, here I have only testing on three datasets to set up a workflow for soil and litter.

- `docs/validation_database.md` is the main documentation, which you should probably start this review with
- `analysis/soil/validation/safe_database_screen/dataset_screening.qmd` is the interactive table Quarto markdown report
- `data/derived/soil/validation/config/sources/` contains a consolidated YAML metadata / configs for the datasets. I explained in the documentation why I recommend them to be tracked and committed like codes to GitHub.
- `tools/R/valdb.R` is the powerhouse function behind this workflow

The rest of the files are either renaming, moving folders, gitignore and various settings. Please let me know your thoughts, thanks!

> [!NOTE]
> The harmonished database is a parquet format not currently committed to Github, since it is a binary. I am happy to show the csv version if it helps to illustrate the final product. I am exploring parquet because I expect the final dataset to be somewhat large sized. Not settled between parquet vs. csv yet.


data/derived/soil/validation/config/sources/10-5281-ZENODO-1198460.yaml
jacobcook1995, line 1: Is there a way to combine sources into a single file? One file per source feels messy and hard to track

data/derived/soil/validation/config/derived_variables.toml
```diff
@@ -0,0 +1,17 @@
+[[variable]]
+name = "total_soil_n"
+description = "Total soil nitrogen"
+unit = "kg N m^-3"
+transform = "soil_cnp_pool_lmwc + soil_cnp_pool_maom + soil_cnp_pool_necromass + soil_cnp_pool_pom + soil_n_pool_ammonium + soil_n_pool_nitrate"
```
jacobcook1995: This should actually also include microbial N for completeness, bit of a pain to extract as you have to divide the biomass of each group by the stoichiometric ratio

I do wonder if it might be worth writing a file of transformation functions, and then referring to them here rather than trying to write them out here (because there will be a hard limit to how complex a transformation you can express inline)

data/derived/soil/validation/config/derived_variables.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "total_soil_c"
+description = "Total soil carbon"
+unit = "kg C m^-3"
+transform = "soil_cnp_pool_lmwc + soil_cnp_pool_soil_cnp_pool_lmwc + soil_cnp_pool_maom + soil_cnp_pool_necromass + soil_cnp_pool_pom + soil_c_pool_arbuscular_mycorrhiza + soil_c_pool_bacteria + soil_c_pool_ectomycorrhiza + soil_c_pool_saprotrophic_fungi"
+
+[[variable]]
+name = "total_soil_p"
+description = "Total soil phosphorous"
+unit = "kg P m^-3"
+transform = "soil_cnp_pool_lmwc + soil_cnp_pool_maom + soil_cnp_pool_necromass + soil_cnp_pool_pom + soil_p_pool_labile + soil_p_pool_primary + soil_p_pool_secondary"
```
jacobcook1995: Same comment as for total N

data/derived/soil/validation/config/unit_conversions.csv
```diff
@@ -0,0 +1,13 @@
...
+ug cm^-3,kg P m^-3,x/1000
+mg cm^-3,kg C m^-3,x
+mg cm^-3,kg N m^-3,x
+mg cm^-3,kg P m^-3,x
+g cm^-3,kg C m^-3,x*1000
+g cm^-3,kg N m^-3,x*1000
+g cm^-3,kg P m^-3,x*1000
+g g{bulk soil SAFE}^-1,kg N m^-3,x*0.9421*1000
+g g{bulk soil SAFE}^-1,kg C m^-3,x*0.9421*1000
+ug g{bulk soil SAFE}^-1,kg P m^-3,x*0.9421/1000
```
jacobcook1995: This is a mix of basic unit conversions and site specific parameter estimates (albeit ones needed for unit conversions). Reckon it's better to keep those things separate if you can, as it removes something that could trip you up when trying to generalise

data/derived/soil/validation/config/units_canonical.yaml
jacobcook1995, line 1: Is there something that you can do with this that you wouldn't be able to do with the `data_variables.toml`, because at a glance the information contained looks identical (just in a different markup language)

tools/R/valdb.R
```diff
@@ -0,0 +1,392 @@
...
+) {
+  # prompt to enter DOI
+  doi <- readline("Enter DOI: ")
+
+  # download dataset metadata
+  meta <- rcrossref::cr_cn(doi, format = "bibentry")
+
+  # set up output path
+  slug <- gsub("[/.]", "-", meta$doi)
+  path <- file.path(outdir, paste0(slug, ".yaml"))
```
jacobcook1995: Think you could instead search for a specific doi entry within a single "sources" file here

jacobcook1995 (commented):
I think the output  of this is useful. I had a few general comments about the workflow, but my knowledge of how you would build something like this in R really isn't sufficient to approve the PR

data/derived/soil/validation/config/sources/10-5281-ZENODO-1198460.yaml
hrlai: There certainly is! I have been juggling between one YAML per dataset vs. one YAML for all datasets. I went for the former because it was easier to `add_schema()` template to each YAML config, whereas the latter would require a modified `add_schema()` to first locate the dataset in the YAML, and then append the template at the right place. It isn't impossible, and I've an idea of how to tackle it.

data/derived/soil/validation/config/derived_variables.toml
hrlai: Let's see what other thinks in https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/1630

data/derived/soil/validation/config/unit_conversions.csv
hrlai: This is another reason for https://github.com/ImperialCollegeLondon/ve_data_science/issues/346, and I wasn't able to put this reason down in words while writing that Issue. If we convert the VE data variables first using VE's bulk density (once we switch back to `abiotic` from `abiotic_simple`, as noted in #343), then we do not need that `g{bulk soil SAFE}` in this unit conversion table.

data/derived/soil/validation/config/units_canonical.yaml
hrlai: No 😅 and I have been anticipating this coming! Yeah, I should totally consider bypassing the need of this YAML file.

tools/R/valdb.R
hrlai: Sorry not sure what you meant here, could you elaborate? Thanks.

jacobcook1995: ahh sorry, basically if you wanted everything in a single sources file, you could always use the same path here, and on the next step check if a particular doi exists within the single file

data/derived/soil/validation/config/units_canonical.yaml
hrlai: This step has been removed!

data/derived/soil/validation/config/unit_conversions.csv
hrlai: I have removed the dependency on site specific values for easier maintenance as suggested. It is stictly speaking still a placeholder for now, but it paves the way for what we'll tackle in #346

data/derived/soil/validation/config/derived_variables.toml
hrlai: Looking back I think I better understand what you meant @jacobcook1995 . Did you mean to replace the lengthy line `transform = "soil_cnp_pool_lmwc + soil_cnp_pool_maom + soil_cnp_pool_necromass +..."` with `function = "convert_soil_n_pools_to_total_n"` ?

I think this is great, but this requires almost one function for each derived variable. I was thinking of an alternative which is more general, like `calculate_derived_variables(..., transform = <the lengthy equation here>)` but this has its cons as well (e.g., it's harder to generalise the `...` part.)

We'll think about it and it'd be great to get some thoughts from you guys too.

@Baizurah and @nickwctan also tagging you two here for discussion later.

jacobcook1995: > Looking back I think I better understand what you meant @jacobcook1995 . Did you mean to replace the lengthy line `transform = "soil_cnp_pool_lmwc + soil_cnp_pool_maom + soil_cnp_pool_necromass +..."` with `function = "convert_soil_n_pools_to_total_n"` ?

Yep this is what I meant, though I think you could also use notation like `convert_soil_pool_to_total_nutrient(nutrient=nitrogen)` where functions are a big more generic.

My concern basically is that `toml` is not a particular readable format (it works if your equations are just simple additions but beyond that it becomes pretty unreadable). So I thought it was better to provide function names, with these functions being written out in proper code (i.e. R/python) with docstrings

davidorme: It's very hard to generalise functions like this - I've tried using symbolic representations of functions before - and it always breaks when it hits reality. Also your functions as encoded in TOML are language specific.

If you are providing a library of derived variable functions, then the data in this TOML is perhaps better recorded as function docstrings.

With these simple addition versions, we might be able to reduce code boilerplate using some kind of template (like an ABC, but for a function) but I suspect that is a marginal gain.

hrlai: I am switching from symbolic representation to bespoke functions now for each target derived variables. Like you said, trying to generalise them feels like a marginal gain, and when I actually put it down in action they aren't that repetitious.

annarallings (approved):
Thanks for going over this with me last week - I think its a great tool and although I can't give detailed technical feedback, let me know how it continues to work for you.

PR #344, opened 2026-05-26
316 Adding new plant variables fruit/seed

### LINKED ISSUE #316: Add new plant input variables
Some additional plant input variables are required:
- `c_mass_fruit_flesh`
- `c_mass_per_fruit_seed`
- `seeds_per_fruit`

The first two are used to calculate flesh to seed mass fraction, so their units can be g or kg (they just need to match).
Seeds per fruit is a count.

These variables will be part of the `pft_definitions` so they can be PFT specific.

Note that fruit and seeds currently share the same stoichiometric ratios (their C pool can change, but related N and P are driven by the reproductive tissue stoichiometry). Therefore, double check the preparation of this stoichiometric input data with regards to how it captures average reproductive tissue stoichiometry.

Update scripts, derived/scenario data in Globus.

arne-exe:
See #316 for details.


arne-exe:
The relevant code from this PR is used on the plant-input-organisation branch instead.

PR #365, opened 2026-06-26
364 update gitignore

hrlai:
This PR cleans up .gitignore and prepares the Maliau scenario input data change log to *not* be ignored. It also does some housekeeping for #103 

@nickwctan I checked your suggestion and it seems that negate `!` is indeed what we needed.

@lelavathy this assumes that the change log will be named `data/scenarios/maliau/maliau_1/change_log.txt` and similarly in Maliau 2.
`


.gitignore
```diff
@@ -492,7 +503,7 @@ FodyWeavers.xsd
 # .vscode/*
 !.vscode/settings.json
 !.vscode/tasks.json
-!.vscode/launch.json
+#!.vscode/launch.json
```
nickwctan, line 506: shall we just delete this if not in use or is there a reason to leave it in there?

nickwctan (approved):
Thanks for making the changes! Looks good, Im glad my suggestion helped. I hope it worked.

.gitignore
hrlai: My Positron is adding settings that I don't want to overwrite your VScode. But I'd like to leave it in as a comment for posterity/as a reminder.

PR #366, opened 2026-06-27
346 functions to compute derived variables

hrlai:
This PR focuses on soil and:
1. adds some functions to calculate derived variables (e.g., total soil carbon) in `tools/R/get_derived_variables.R`
2. adds unit tests for these functions in `tests/testthat/`
3. adds a [vignette](https://github.com/ImperialCollegeLondon/ve_data_science/blob/54ea5083d59a6442b8ca950adf0321230c06b2bb/docs/derived_variables.md) in `docs/derived_variables.md`

@jacobcook1995 could you review features 1-3? Or just 2, since unit test is the major thing we discussed.

@annarallings could you review feature 3?

@arne-exe could you review feature 1, after getting a sense from the vignette?

@sallymatson could you review feature 2 and 3? These are R stuff, but I'm just looking for a high-level feedback from you, if any (e.g., I hope `tests/` is compatible with `tests/pytest/` in case we set the latter up one day). Then, there is also `tools/python/generate_config_test.py` that we talked about.

@nickwctan @Baizurah and @lelavathy tagging you here for the single python script, and the overall gist to keep you posted. Thanks all!


tools/python/generate_test_config.py
```diff
@@ -0,0 +1,36 @@
+"""Module to generate test configurations for the virtual ecosystem."""
```
nickwctan: I think I see what you are trying to do here - creating a config toml file in your mock file path for further testing purposes...?

May i suggest an alternative naming? Would `generate_config_for_testing.py` be better?

nickwctan (changes requested):
Feature 1 looks good. Can't comment much on Feature 2 at the moment. Just had a suggestion on the naming on your py file to ensure that it is not confused as a pytest file.

tools/python/generate_test_config.py
hrlai: Yes! Thanks, I wasn't aware that pytest naming convention also includes `*_test.py`, good to know. I'll change the filename!

annarallings (approved):

nickwctan (approved):
LGTM

jacobcook1995 (approved):
This all looks good to me, gives us a great starting point that we can build on for more complex derived variables!

In terms of the testing I'm not familiar with `testthat` but the usage looks sensible. The reason I suggested adding unit tests was so that if we ever want to port the functions to calculate derived variables to python we have expected behaviour that they need to match, which this PR provides

I'm not sure how you are currently running the unit tests but I would suggest adding them as part of the CI workflow so that the tests are automatically run on each and every PR. This is [how they are run on the Virtual Ecosystem repo](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/06fa64b20136510b558219292c17d2b94869e336/.github/workflows/ci.yml#L46-L47). Setting up `testthat` unit tests will be similar but not identical (happy to help if you run into any difficulties!)

PR #371, opened 2026-07-02
Demo for reading/writing Zarr outputs in R

hrlai:
This PR demonstrates examining and manipulating VE output stored in Zarr format using `pizzarr` in R. It also shows how to convert Zarr to netCDF by piggy-backing on `xarray` from Python in R. @annarallings hopefully this addresses your question in https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1667#issuecomment-4843417230

There is also a setup script in bash to setup virtual environment, install the dev VE version, install example data, and then run `ve_run` to get the new Zarr output files.

This PR is intended to develop in parallel with another PR (https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1667) on the VE repo.

@arne-exe this PR is probably best reviewed by you, to make sure that the steps in `demo.R` make sense. It would be good to hear from you if switch to Zarr from NetCDF is okay for R users. Overall, the very quick jump from Zarr to R array class feels very good, but there will be a bunch of functions / scripts that need revamping (or can we leave them as legacy?), so there will still be a fair bit of overheads at the start, until we slowly get our return from investment in the form of much faster model compilation (can't tell how much until some profiling is done).

Last but not least, @annarallings mentioned in https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1667#issuecomment-4843417230 that:
> So the input data will still be NetCDF (where applicable), but all the outputs that were NetCDF will now be zarr, correct?

I think this is something worth discussing, but probably in another thread?


davidorme (approved):
LGTM

I can't quite get used to R6 classes in R. Something about `Instance.do_a_thing()` looks wrong when it is `instance$do_at_thing()` 😄

analysis/troubleshoot/zarr_test/demo.R
```diff
@@ -0,0 +1,157 @@
...
+# Open the model output Zarr store
+zarr_path <- "analysis/troubleshoot/zarr_test/ve_example/out/model_data.zarr"
+outputs <- zarr_open(zarr_path)
+
+
+# Examine and manipulate Zarr file ---------------------------------------
+
+# Print the store hierarchy and inspect the top-level groups
+# the print output can be long given the large number of VE output variables
+outputs$tree()
```
arne-exe, line 70:
```
> outputs$tree()
/ 
├── init
Error in initialize(...) : 
  blosc package must be installed to use the Blosc codec. Install with install.packages('blosc')
```

arne-exe (approved):
Using `model_data.zarr.zip` I was able to run the demo script.
It all seems to be working well, very useful demo!

PR #384, opened 2026-07-08
Migrate to uv for dependency management and add how-to documentation

hrlai:
This PR:
- migrates the Python environment and dependency management from `poetry` to `uv`
- updates [documentation ](https://github.com/ImperialCollegeLondon/ve_data_science/blob/uv/docs/uv_setup.md)to reflect this change and setup instructions
- replaces the `pyproject.toml` configuration to use `uv`

My motivation is to make it as easy to run multiple versions of `ve_run` as possible. The goal is to make even first-time python users feel confident and painless. Currently it is making a positive QoL improvement to my workflow in PR #329 where I had to keep track of the dev versions. 

And I aim to only set up the baseline `uv`. Python coders can slowly add dependencies that suit in future PRs (for example, you can add `numpy` or `matplotlib` later; this should also make sure that we can run each other's code?) For R users, this is probably close to the minimal that we ever need, i.e., 90% of the time our main use is to easily switch between developmental and release version of VE. 

@davidorme and/or @sallymatson , could we get your feedback on the `uv` files, and especially the [tutorial ](https://github.com/ImperialCollegeLondon/ve_data_science/blob/uv/docs/uv_setup.md)`docs/uv_setup.md`? Thanks!

p/s: I have nothing against `poetry` --- I actually never used it for our projects. I am exploring `uv` simply because it was recommended to me by a friend, and skimming their documentation it seems user friendly enough for someone like me. Let me know how you feel about the migration?


docs/uv_setup.md
```diff
@@ -0,0 +1,187 @@
...
+### macOS / Linux
+
+```sh
+curl -LsSf https://astral.sh/uv/install.sh | sh
+```
+
+Check that it works:
+
+```sh
+uv --version
```
arne-exe, line 45:
```suggestion
uv --version

If this doesn't work, make sure that uv's install location is added to PATH.
```

docs/uv_setup.md
```diff
@@ -0,0 +1,187 @@
...
+
+```sh
+uv sync
+uv sync --group dev
+uv sync --group dev-pinned
+```
+
+Only one version is active at a time, and each command switches to that version.
+Simply switching groups does not need any extra flags; `uv sync` will install
+whatever version that group needs. See
```
arne-exe, line 89: What's the recommended approach to see the actual code being used?
Going into `ve_data_science\.venv\Lib\site-packages\virtual_ecosystem\models`?
Does the VE code in this folder switch when switching between `uv sync --group dev` and `uv sync --group dev-pinned`?

docs/uv_setup.md
```diff
@@ -0,0 +1,187 @@
...
+
+Run:
+
+```sh
+uv lock --upgrade-package virtual-ecosystem
+uv sync
+```
+
+This updates `uv.lock` (the file that records exact package versions), then
+installs the newest compatible PyPI release of `virtual-ecosystem`.
```
arne-exe, line 107: Since the file is tracked by Git, should we ignore the tracked message or should we push the update with a commit?

docs/uv_setup.md
```diff
@@ -0,0 +1,187 @@
...
+
+Run commands or `.py` scripts with `uv run`. For example:
+
+```sh
+uv run ve_run ...
+uv run my_script.py
+```
+
+> **Warning:** If you synced with `--group dev` or `--group dev-pinned`, pass
+> the same group to `uv run` to avoid switching back to the default:
```
arne-exe: Perhaps rephrase to something easier to understand like "If you want to run the VE using the develop version, then..". I understand syncing refers to the "old switching virtual environments" but it's not super intuitive, so rephrasing (or extra clarification) may help for the tutorial here.

arne-exe (approved):
I've added a few question/suggestions for clarification, but nothing critical, so will approve. Overall the tutorial is clear and the install process seems clearer and less confusing for a new user.

docs/uv_setup.md
hrlai: Yes, exactly how you described! I confimed that switching between versions, the VE codebase also changes. 

@nickwctan tagging you because we discussed how to feed this into Copilot.

docs/uv_setup.md
hrlai: If I understood correctly, you are asking if `uv.lock` should be tracked and commited to Git?

docs/uv_setup.md
hrlai: Thanks for pointing out, I incorporated your suggestion.

docs/uv_setup.md
arne-exe: I phrased my question poorly. It looked like `uv.lock` is already being tracked, because when I ran `uv lock --upgrade-package virtual-ecosystem` and `uv sync` this changed `uv.lock` which led to Git showing it in the list of changed files. My question is more related to if this is supposed to happen, and how/when `uv.lock` is expected to change when we switch between commits. For example, if I want to do some testing by running the VE 3 times, each time from a different commit. What would the process look like and which files would change?

PR #407, opened 2026-07-16
Notebook folder reorganisation

hrlai:
This PR move the notebook tutorial folder by @sallymatson into the `templates` directory to reduce clutter, and that also seems like a natural home for it.

There are a Jupyter notebook template for R by @nickwctan , and a myst markdown template by @davidorme , since a while back. They are a bit duplicated but I think it is okay to leave them as is, so I did not touch them.


sallymatson (approved):
This reorg makes sense to me!

PR #411, opened 2026-07-16
Update soil layer depth terminology in derived variable calculations

hrlai:
This PR updates the R tools and tests to use the `microbial_simulation_depth` constant parameter instead of the deprecated `max_depth_of_microbial_activity` for soil layer depth calculations, matching https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1688


jacobcook1995 (approved):
LGTM!

PR #412, opened 2026-07-16
Experimental ast-based docstring extraction and constant--function mapping

hrlai:
This PR aims to communicate with @jacobcook1995 what I need and ask for code review. To tackle #403, for a constant in VE, I need to know at least three things:
- [X] Its unit
- [X] Its description
- [ ] The functions that used it, and the docstring of those functions

I want them in a table with columns:
- constant
- unit
- description
- function
- function_doc

The first two I can get from the constant's Pydantic config. But the third is trickier, and I'm trying to parse the source codes using ast. The aim is to give the context about the constant in a function or equation. For example, is constant A used in a linear or non-linear function, is it a proportional change, is it a conversion factor, does it have an alias or alternative name used in the literature etc. Ultimately this is to give the LLM enough context to look for the right empirical value.

Some challenges I need to solve are:
- a constant can be called by multiple functions
- and those functions may call it via another nested function

In the latter case, I expect the long-format output table to have multiple columns for a constant, each row mapped to a function that calls it.

The single Python script in this PR is drafted by me and Copilot. I wanted something more concrete then the rambling above. I tried to make it clear with is the user-facing function (i.e., `extract_constant_call_doc_map()`) and documented it with a crappy example (see its docstring). 

@jacobcook1995 could you give this a skim and let me know if ast is the way to go?

I do not expect this script to be the final tool, and I expect us to rewrite it even from scratch. It is here mainly as a proof-of-concept only. Thanks!


analysis/soil/llm/constant_call_doc_extractor.py
jacobcook1995: Would add a comment to explain this regex, I guess it also catches "Raises:" and "Args:" which I guess you want

jacobcook1995: Same comment here

jacobcook1995 (approved):
I had a couple of minor stylistic comments, but I didn't have much other feedback as `ast` is not something I'm familiar with.

I think as this is an experiment it's probably worth merging and seeing how the results come out, as that will (a) identify problems/bugs with the code and (b) let us know how viable the approach seems overall

analysis/soil/llm/constant_call_doc_extractor.py
jacobcook1995: Would be best to add an Args + Returns block to this docstring

jacobcook1995: Again, best practice to include an Args and Returns block

jacobcook1995: Args + Returns

jacobcook1995: Args + Returns

jacobcook1995: Args + Returns

jacobcook1995: Args + Returns

jacobcook1995: Args + Returns

jacobcook1995 (commented):
Also this would be my main code quality suggestion, sorry for the same comment 7 times!

analysis/soil/llm/jedi_constant_usage_tool.py
```diff
@@ -0,0 +1,203 @@
...
+            #       continue
+            #   bases_match = re.search(r'class\s+\w+\((.*?)\):', name.get_line_code())
+            #   if not (bases_match and 'Configuration' in bases_match.group(1)):
+            #       continue
+            #
+            # Benefits: (1) Filters by type first (more efficient), (2) Properly parses
+            # base classes from the class definition line, (3) More robust than string
+            # matching and clearer about intent, (4) Handles different formatting gracefully.
+            if "(Configuration)" not in name.get_line_code():
+                continue
```
davidorme: You could put the type check in the main `if` test - I'm not sure it's a huge speed gain as the string check is so cheap.

I see where you're going with the regex  - and I do [love a regex](https://github.com/davidorme/band_equipment_optimiser/blob/a877fef28ec13131e9609c5b1fb79fa410999e9d/band_equipment_optimiser/parser.py#L123-L138) - but honestly I think it is overkill for a simple string match

The one possible edge case would be:

```python
class ConstantsClassWithAReallyReallyReallyReallyReallyReallyReallyLongName(
    Configuration
):
```

but lets cross that bridge if it ever occurs!

```suggestion
            # TODO: Consider using regex to parse base classes from name.get_line_code()
            # instead of naive string matching. Example:
            #
            #   import re
            #   if name.type != 'class':
            #       continue
            #   bases_match = re.search(r'class\s+\w+\((.*?)\):', name.get_line_code())
            #   if not (bases_match and 'Configuration' in bases_match.group(1)):
            #       continue
            #
            # Benefits: (1) Filters by type first (more efficient), (2) Properly parses
            # base classes from the class definition line, (3) More robust than string
            # matching and clearer about intent, (4) Handles different formatting gracefully.
            if (name.type != 'class') or ("(Configuration)" not in name.get_line_code()):
                continue
```

davidorme (approved):
LGTM.

analysis/soil/llm/jedi_constant_usage_tool.py
hrlai: Add a type check for `name`

PR #413, opened 2026-07-17
Demo for reading/manipulating Zarr outputs in python

nickwctan:
This PR demonstrates examining and manipulating VE output stored in Zarr format using xarray (with zarr dependencies installed) in python. In particular, it demonstrates the use of `xr.open_dataset()` to access zarr files - see  https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1667#issuecomment-4845117216

I included a setup script for linux and macOS to setup virtual environment, install the dev VE version, install example data, and then run ve_run to get the new Zarr output files, following 'setup.sh'. The setup won't be effective when we switch to 'uv' #384 

Perhaps @lelavathy and @Baizurah could review this to see if it makes sense. Feel free to make any suggestions or additions.

### DIFF
--- a/analysis/troubleshoot/zarr_test/demo.py
+++ b/analysis/troubleshoot/zarr_test/demo.py
@@ -1,0 +1,146 @@
+"""
+---
+title: Demo for working with output Zarr files in python
+
+description: |
+    This demo shows how to read and work with Zarr files
+    produced by the virtual ecosystem model. It demonstrates how to access the data,
+    log transform it, and save a slice of the data to a new NetCDF file for further
+    analysis.
+
+virtual_ecosystem_module: All
+
+author:
+  - name: Nick
+
+status: final
+
+input_files:
+  - name: model_data.zarr
+    path: analysis/troubleshoot/zarr_test/ve_example/out/
+    description: |
+      VE model output in Zarr format, containing inputs, init conditions, and outputs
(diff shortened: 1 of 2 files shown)


analysis/troubleshoot/zarr_test/demo.py
```diff
@@ -0,0 +1,146 @@
...
+# ============================================================
+# List inputs and outputs
+# ============================================================
+
+inputs = t["inputs"]
+inputs
+
+outputs = t["outputs"]
+outputs
+
```
lelavathy, line 100: print("Available output variables:") 
print(list(outputs.data_vars))

This gives users a quick overview of what is stored in the outputs group and makes it easier to choose variables for further exploration or analysis.

lelavathy (approved):
Overall, I think this is a good introductory example for working with VE outputs stored in Zarr format. I tried the demo locally and it worked.

analysis/troubleshoot/zarr_test/demo.py
nickwctan: Great idea! Thanks!

Baizurah (approved):
LGTM for starter. From my understanding, the idea is to have script to translate ZARR output according to what we (user) need, which means we can eventually custom made it. Thanks Nick

PR #414, opened 2026-07-20
Upload refactored ERA5 climate input data preparation script

lelavathy:
## Summary

This PR refactors the ERA5 climate input preparation workflow into reusable modules while preserving the existing functionality.

### New modules

- `tools/python/abiotic/cdsapi_downloader.py`
  - Downloads ERA5-Land monthly averaged datasets
  - Downloads ERA5-Land hourly 2 m temperature
  - Automatically extracts downloaded archives

- `tools/python/abiotic/climate_tools.py`
  - Processes downloaded ERA5-Land datasets
  - Calculates monthly mean diurnal temperature range (DTR)
  - Performs spatial interpolation to the Virtual Ecosystem grid
  - Converts variables into VE-compatible format
  - Generates the final climate dataset

### Driver script

- Added `analysis/abiotic/era5_climate_data/maliau_climate_data_processing_script.py`
  - Provides a complete workflow for downloading and processing ERA5 climate inputs.

### Processing improvements

- Corrected radiation conversion from J m⁻² to W m⁻² using the appropriate daily conversion factor.
- Added monthly mean diurnal temperature range derived from hourly ERA5-Land 2 m temperature.

## Testing
- Successfully generated the Maliau climate dataset using the refactored workflow.
Both the climate input datasets for maliau_1 and maliau_2 have been uploaded to Globus. Change log also updated.

Fixes #415

### DIFF
--- a/analysis/abiotic/era5_climate_data/maliau_climate_data_processing_script.py
+++ b/analysis/abiotic/era5_climate_data/maliau_climate_data_processing_script.py
@@ -126,9 +126,11 @@ references: |
 ---
 """  # noqa: D212, D205
 
-import importlib
 import sys
 from pathlib import Path
+
+from tools.python.abiotic import cdsapi_downloader as cds
+from tools.python.abiotic import climate_tools as ct
 
 # ============================================================
 # PROJECT ROOT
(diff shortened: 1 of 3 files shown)


annarallings (approved):
LGTM

analysis/abiotic/era5_climate_data/maliau_climate_data_processing_script.py
```diff
@@ -0,0 +1,508 @@
...
+)
+cdsapi_era5_monthly_downloader = cdsapi_downloader.cdsapi_era5_monthly_downloader
+
+# ============================================================
+# IMPORT CLIMATE PROCESSING TOOLS
+# ============================================================
+# Import functions for processing, interpolating, and
+# formatting ERA5-Land climate data into a Virtual Ecosystem
+# climate forcing dataset.
+
```
vgro: This section is fine but you probably don't need to import all the functions individually. You could just `import climate_tools` or `import climate tools as ct` if you wanted it shorted and then call the functions as `climate_tools.add_atmospheric_co2` or `ct.add_atmospheric_co2`. No need to change now but something to consider in the future :-)

analysis/abiotic/era5_climate_data/maliau_climate_data_processing_script.py
```diff
@@ -0,0 +1,508 @@
...
+save_dataset = climate_tools.save_dataset
+select_required_variables = climate_tools.select_required_variables
+
+
+# ============================================================
+# USER SETTINGS
+# ============================================================
+# Configure the scenario and bounding box for ERA5-Land climate data download.
+
+
```
vgro, line 155: Can the info in this section be derived from the grid definition file? Just to avoid hardcoded changes where possible

tools/python/abiotic/cdsapi_downloader.py
```diff
@@ -0,0 +1,384 @@
...
+def cdsapi_era5_monthly_downloader(
+    years,
+    bbox,
+    outfile,
+):
+    """Download ERA5-Land monthly averaged reanalysis.
+
+    Variables downloaded
+    --------------------
+    - 2 m temperature
```
vgro: Could you please add the units here

vgro (approved):
LGTM, a few minor comments.

analysis/abiotic/era5_climate_data/maliau_climate_data_processing_script.py
lelavathy: Thanks, that's a good suggestion. I originally tried deriving the download bounding box from the grid definition, but the grid represents the VE model/interpolation domain (~1 km × 1 km), which is too small for ERA5-Land (~9 km resolution) and results in a MultiAdaptorNoDataError from CDS. I agree that hard-coding the download extent isn't ideal, so a good future improvement would be to add the details of the ERA5-Land download bounding box to the TOML file.

PR #421, opened 2026-07-20
Reorganise tools folder

nickwctan:
## Summary

Reorganise the `tools/` directory to make the Python and R tooling easier to navigate and extend.

@hrlai happy for you to co-author this PR. 

### Python
- add a package-style layout under `tools/python/`
- place Python source code under `src/`
- keep Python tests under `tools/python/tests/`
- Add minimal `pyproject.toml`  to support pytest config (and not packaging and  dependency management _yet_)

### R
- keep R functions under `tools/R/R/`
- keep tests under `tools/R/tests/testthat/`

## Why

This creates a clearer separation between Python and R tooling and gives both a more proto-package structure, which should make development, testing, and future packaging easier.

## TODO

- [x] Move `tools/python/animal/test_trophic_mass_flow.py` → `tools/python/tests/
test_trophic_mass_flow.py`
- [x] Create `tests/python/conftest.py` for shared test helpers
- [x] update any relative paths/imports in the moved test
- [x] Add `pyproject.toml`
- [x] Add pytest configuration to `pyproject.toml` if needed
- [x] check CI still picks up both Python and R tests

Fixes #408


hrlai (approved):
LGTM on the R side. I have made several minor updates to the paths in scripts and docs. The R tests will not pass (also the same for other recent PRs) because the config generator depends on VE, which is experiencing pyrealm issues, and I'll revisit it once we have another VE release. 

Thanks for this!

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install test dependencies
```
hrlai, line 36: Looks good to me, except maybe add a note or Issue to update this step when we move to `uv`, but do that later. I have a placeholder step in `.github/workflows/r-tests.yml`

tools/python/pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+addopts = "-ra"
+
+[tool.ruff]
```
hrlai: Do we need this when it is already in the root `pyproject.toml`?

hrlai (approved):
Looks good to me, with minor questions. I can't do a hard review on `conftest.py` but I trust that it ran well on your side.

davidorme (approved):
Looks good.

The only thing I'd be wary of here is if the CI testing gets in the way of the day to day business of the repo. Mostly you aren't changing the tools, you're changing scripts and metadata etc. But these CI tasks run on all PRs. They are probably pretty fast but the setup of the test environment can take some time and waiting for tasks to pass before being able to merge _could_ end up being a real day to day irritant for the DS team.

I'm not saying don't do this _at all_ because I think getting testing up and running is important, but I am saying be wary of how building out this part of the repo can impact your main interactions with it and keep an eye on how much friction that is causing (@annarallings - something to add to your watchlist 😄). 

The long term answer of course is to move them to different repos and make them into real packages, because then the CI runs only when there are tool code changes.

tools/python/pyproject.toml
nickwctan: Good catch. no we dont :)

.github/workflows/python-tests.yml
nickwctan: Yes I added to a reference to #384 for the subsequent steps. Thanks!

PR #422, opened 2026-07-21
Pin mkdocs-material version to 9.7.0

hrlai:
This pull request updates the deployment workflow to ensure a specific version of the `mkdocs-material` package is used during deployment. The pinned version is based on the first version that `mkdocs-material` went into [maintenance mode](https://github.com/squidfunk/mkdocs-material/releases#release-9.7.0).


jacobcook1995 (changes requested):
I'm not 100% sure how this works with `uv`, but with `poetry` the main place you want to pin python package versions is the `pyproject.toml` file

PR #431, opened 2026-07-23
Progress report on LLM literature search for constant values

hrlai:
> [!IMPORTANT]
> **tldr:** Take a look only at the [LLM prompt](https://github.com/ImperialCollegeLondon/ve_data_science/blob/fa75fa94e591237c3e392aabda1dccd059f718c9/analysis/soil/llm/chat.R#L92-L172) if you're interested.
>
> Sorry for tagging everyone to review this! First of all, you don't have to do this, but I do not know who to request review from because (1) I'm not sure who else does LLM programmatically in R, (2) even if you do you'll have to set up the API and access keys to the LLM, and (3) you'll have to have installed VE in the same folder as me (the recent `uv` merge can make this easier, but still).
>
> This scoping exercise is phased and this represents the completion of step 1: provide Rob a crude sense and dig out problems. I could continue working on this branch, or merge it into a future larger branch, but that would grow out of control. I think this is a natural checkpoint to merge because the code "does nothing", they are proof-of-concept code get me to a failure point (insufficient token on free account) to report back, but they represent a significant amount of exploration.
>
> This however means that it is hard to know where to start reviewing. So I'm suggesting:
> 1. Read (and review) the [LLM prompt](https://github.com/ImperialCollegeLondon/ve_data_science/blob/fa75fa94e591237c3e392aabda1dccd059f718c9/analysis/soil/llm/chat.R#L92-L172) in `analysis/soil/llm/chat.R`
> 2. Skim through the list of information that an online search should retrieve, they are specified [here](https://github.com/ImperialCollegeLondon/ve_data_science/blob/fa75fa94e591237c3e392aabda1dccd059f718c9/analysis/soil/llm/chat.R#L174-L228). They are table columns like searched values, unit, publication DOI/URL, confidence level etc.
> 3. There is also the report I wrote to Rob `analysis/soil/llm/scoping_report.md` (_or read this first_)

This PR adds:

1. R scripts to explore the feasibility of LLM to search for constant values
    - `analysis/soil/llm/rag.R` builds a retrieval-augmented generation (RAG) store that grounds the next script with VE codebase information
    - `analysis/soil/llm/chat.R` uses the RAG to understand the VE codebase, then search online for soil constant values
2. The key file is `analysis/soil/llm/scoping_report.md` to report back to Rob

Two deprecated files not worth detail reviews:
- `tools/python/src/ve_data_tools/constant_usage_tool.py` is a static code analysis tool to parse VE codebase courtesy of @davidorme , but we have decided that it does not work in #423 that this route does not work. I have turned to the RAG approach, for now. Being able to directly parse the codebase is still a better source of truth, but that is too time consuming for our purpose now (which is just to scope if LLM is feasible)
- `analysis/soil/llm/extract_soil_constant_metadata.R` is my R script sourcing the above deprecated Python tool to do some downstream analysis, also deprecated

I am leaving these two deprecated files in this PR, in case they are worth revisiting.


analysis/soil/llm/extract_soil_constant_metadata.R
```diff
@@ -0,0 +1,110 @@
...
+use_virtualenv(here(".venv"), required = TRUE)
+cu <- import_from_path(
+  "constant_usage_tool",
+  path = "tools/python/src/ve_data_tools"
+)
+
+
+# Map VE constants to the caller functions -------------------------------
+
+# A list of model files containing constants to parse
```
jacobcook1995, line 75: I guess all of the constants exist in `model_config.py`. What is the value in feeding in all the other files as well? Is it a case of giving more context to the LLM?

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -0,0 +1,193 @@
...
+    project_root=None,
+):
+    """Find usages of configuration constants in target file(s) and write to TOML.
+
+    Scans each target file for classes that inherit from ``Configuration``,
+    collects every attribute defined on those classes, and uses ``jedi`` to find
+    all reference sites across the project.  All results are merged into a single
+    TOML file at ``out_path``.
+
+    Parameters
```
jacobcook1995: This is minor but it is might be best to use [google docstrings style](https://google.github.io/styleguide/pyguide.html) here as that is the style we use in `virtual_ecosystem`. You've captured all the important information, it just requires a few minor formatting changes

jacobcook1995 (commented):
This was really interesting to read! I don't have much to add honestly as this is substantively outside my expertise.

(I made a couple of minor comments, but I'm not sure of their relevance given you have decided on the RAG approach)

davidorme (approved):
Looks interesting. As with @jacobcook1995 this is not something I'm hugely familiar with, but happy for this to go in.

I think we can use the `jedi` outputs in the same way that you are using RAG? As I understand it, RAG is providing a database of dismantled chunks of code that help the model localise where constants are being used. The `jedi` output enable the same basic localisation: we can tell an LLM in the `chat.R` prompt that it should look at "these references in the `virtual_ecosystem` codebase" where each constant is used.

The only failing of `jedi` here is that sometimes the constant is passed down to a function by value rather than name, so we lack a reference for the final calculation. However, it isn't clear to me that using RAG allows an LLM to "understand" passing by value, so it is also unclear to me that it actually fixes the `jedi` problem? We could take care to prompt the LLM to look at the explicit references but also follow calls within that reference where the constant value as an input.

analysis/soil/llm/extract_soil_constant_metadata.R
hrlai: I guess I wanted to not only feed context about the constant definitions, but also their caller functions, which contains the functional forms, equations etc. Does that make sense?

tools/python/src/ve_data_tools/constant_usage_tool.py
hrlai: Updated, thanks!

analysis/soil/llm/extract_soil_constant_metadata.R
jacobcook1995: ahh yes it does!

PR #457, opened 2026-07-30
Enhance R script template to facilitate the newly added AGENTS.md

hrlai:
This PR aims to use Copilot to populate a draft metadata header for every new R scripts, let the user reviews it. This will save a lot of time on documentation. 

However, if we use Copilot in IDE as is, it tends to write the metadata from scratch, rather than depending on the existing templates in the `templates` folder. A solution is to tell it to do so in the prompt, every single time. This is a [perfect situation](https://awesome-copilot.github.com/learning-hub/building-custom-agents/) to leverage an AGENTS.md file to overcome such a repetitive task.

What I did:
- Use the [create-agentsmd](https://github.com/github/awesome-copilot/blob/main/skills/create-agentsmd/SKILL.md) skill from Awesome Copilot to generate a template AGENTS.md file
- Use the `/savememory` skill in Positron Assistant to add the section "Script metadata header guidance" to tell the agent how to populate the metadata header using our templates
- Update the R script template to provide a cleaner context for the agent. I did so while keeping the template human friendly (i.e., I did not aim for it to be optimised only for an AI). While doing so, I added the `source_files` section since we have already been using it in some R scripts, but haven't updated it here.

After setting these up, I prompted Copilot in IDE again and it drafted me a near-perfect metadata header.

@arne-exe asking you to review the R template, if the change remains okay.
@davidorme tagging you because you authored the original R metadata template.
@nickwctan tagging you because you have been using Copilot. Hopefully you can take a quick review on the AGENTS.md file too.
@annarallings tagging you because we have been talking about assisted documentation in general.


arne-exe (approved):
Looks good!

PR #459, opened 2026-07-30
Update change log on adding scenario specific plants.constants block

Baizurah:
Added a custom plants.constants block with scenario-specific values.


arne-exe (approved):
Looks good, thanks for adding them back in.

hrlai (approved):

PR #461, opened 2026-07-30
Animal - Document validation plan

### LINKED ISSUE #462: Define animal validation targets
After the meeting with Rob, it would be ideal to have a set of validation targets ready for Rob to review or for us to go through with him. This will facilitate finding site-specific datasets , specifically SAFE datasets which contain Maliau and SAFE data.

nickwctan:
This is a preliminary validation plan - a working [document](https://github.com/ImperialCollegeLondon/ve_data_science/blob/99edaaedcdea52dd05b23732350dc17560093d1e/docs/animal_model_validation_plan.md) with @Baizurah where we can make updates according to our validation progress. In particular, it is to document our validation targets for our next meeting with Rob. 

Feel free to comment or work on this file together, Bai. Would be curious to hear what your opinions are! 

Fixes #462


TaranRallings (approved):
This looks good to me. All of the big bases I can think of are accounted for. It might benefit from some more explicit language about food webs in the trophic structure section but it's not a pressing concern.

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,467 @@
...
+
+Notes:
+
+- Perhaps some of the sections can be split into smaller sections or subsections with lesser variables?
+- Decide whether we want to be already very specific with our validation approaches? Or keep it rather general now to find out more about available validation datasets first? Currently validation appraoches are very broad.
+
+## 1. Scope and rules
+
+1. Animal-model validation only, including cross-module outputs when they directly support animal interpretation.
+2. Validation metrics are computed from exported outputs and post-processing only.
```
Baizurah: How about other target that presently cannot be derived from exported outputs but turned out to be important, which part does it falls into... ie: cause of mortality rates. We have mortality cause by senescence, predation and starvation, but in the exporter currently we only have cohort state (is_alive).

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,467 @@
...
+3. Targets are organised into 1) site-specific emergent or derived targets and 2) global emergent pattern checks (Madingley based).
```
Baizurah: Propose to: Targets are organised into 1) site-specific validation targets, including derived and emergent metrics, and 2) global emergent-pattern checks based on Madingley and other published known ecological benchmarks. 

else, it seems like all site-specific targets must be emergent, which I dont think you mean it?

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,467 @@
...
+
+## 2. Secondary validation targets (emergent or derived from outputs)
```
Baizurah, line 23: I am missing No.1/ Primary validation targets..

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,467 @@
...
+
+Targets in this section are site specific.
+
+### 2.1 Population and biomass density structure by functional group
+
+*Derivation:* Derive post-spin-up population density, biomass density, and density-mass scaling from cohort abundance, occupancy, territory, and elemental mass.
```
Baizurah: @TaranRallings does the “spin-up” here mean the same thing that I asked about before? Or is it just a choice of words to mean that we’re not using the specified density at the initial time, but instead allowing the population to run until it becomes stable? If so, how long does it usually take to stabilise?

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,467 @@
...
+
+*Inputs:* individuals, functional_group, occupancy_proportion, territory_size, mass_carbon, mass_nitrogen, mass_phosphorus, largest_mass_achieved.
```
Baizurah: `population_densities` by FG is available in VE animal_cohorts.py.  Landscape density (density of a functional group across the whole simulated landscape) can also be derived from individuals by FG and total area, but the current `population_densities` still not counting for multi-grid occupancy.  do you think `occupancy_proportion` and `territory_size` perhaps fit better under spatial/territory validation, while `largest_mass_achieved` fits better under growth/maturity? As this infer to largers mass reached by a cohort... not standing biomass..  @nickwctan

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,467 @@
...
+
+*Validation approach:* Compare equilibrium density ranges, biomass density, and density-body-mass scaling against published benchmarks for comparable tropical or temperate forest systems. Stratify by body-mass class and diet class where available.
```
Baizurah: Hmm... at this point, do we know if its going to reach equilibrium... Perhaps we can revisit after running the test after excluding the initial state. 

Density–body-mass scaling: I assume this is intended as an emergent-pattern validation, right. Since VE initial population densities are   pre-determined using body-mass scaling (Damuth/Madingley), so validating this relationship at the initial stage would be circular. It would make more sense to assess whether the density–body-mass relationship is maintained or emerges after the model has been running.. did i understand this correctly... @nickwctan

docs/animal_model_validation_plan.md
nickwctan: Yep, thats what I removed in the last commit - see https://github.com/ImperialCollegeLondon/ve_data_science/pull/461#issuecomment-5237278588

docs/animal_model_validation_plan.md
nickwctan: > , it seems like all site-specific targets must be emergent, which I dont think you mean it?

Hence `emergent or derived targets`. I am not quite sure what you mean. 

> other published known ecological benchmarks.

As of now, I am only using Madingley but if we add other benchmarks, I am happy to add this part in.

docs/animal_model_validation_plan.md
nickwctan: Thats a very good question which I do not have the answer for :D 

I am sure there is a way to calculate it backwards using parameter constants like u_bg, or export it at every update. The question is, do we have that detailed dataset to validate these values against? I'd be quite surprised if there is a field data collection which tease apart causes of mortality together with no. of deaths.

docs/animal_model_validation_plan.md
nickwctan: Populations can never be stable - but can achieve a somewhat equilibrium stage, which I am sure that's what you meant. Yes to running arbitrary initial conditions (especially when we are unsure of the population density), allowing it to run to a (sort-of) equilibrium stage. 

We could use that spin up density values to validate against our data and actually ideally long term datasets which we can validate throughout the simulation months/years, see if we can match the oscillating cycles of population

docs/animal_model_validation_plan.md
nickwctan: Good catch on the density-body-mass scaling! Thanks. Will remove it

docs/animal_model_validation_plan.md
nickwctan: occupancy_proportion and territory_size are both in the spatial/territory validation as well. I am leaning towards leave these variables in here too as there might be a possibility we compare density larger than one cell level. 

largest_mass_achieved is also in growth/maturity. You are right I will remove it from here, replacing it with is_mature as some studies might only look at adults individuals.

docs/animal_model_validation_plan.md
```diff
@@ -0,0 +1,466 @@
+<!-- markdownlint-disable MD013 MD010 -->
```
davidorme: Please don't arbitrarily turn off formatting rules. The whole point of using shared QA tools is to have standard formatting.

nickwctan: Thanks for the kind reminder. I've removed it except for the wide tables which I have left them on so we can deal with them later (e.g. use CSV as registry) should we decide to put this document up on the website.

annarallings (approved):
Great exhaustive document, I appreciate listing out all the variables and their attributes too. That will come in handy. It's definitely a huge list and I guess Rob's assessment will give you an idea of how to initially whittle it down. In each of these categories, you may need to identify the functional groups/areas of coverage you can get with these. I am sure some groups of species will have major gaps. The question will be which ones are most ecologically significant to the food web/community structure and whether we have enough relevant data to address it. Some of these variables may also be "keystone" kinds of variables that really show whether the animal model is working, though Taran and the Madingly work will have some good thoughts on that.

docs/animal_model_validation_plan.md
TaranRallings: Spin up just means a period of running the model before actually running the model to get a "sort of" equilibrium (or maybe an actually stable equilibrium, we'll have to wait and see).

nickwctan:
Thanks all for the review. I have updated the document, narrowing down to a couple of possible validation targets we can achieve.  Priorities will be given for the direct validation targets at the moment. 

I am pretty happy about it now, especially that we have reference datasets for all direct validation targets. So after you guys can have a quick look at this, I can close the PR. Ofcourse this doesn't mean the plan is final. 

@Baizurah @TaranRallings @annarallings  Updated md file [here](https://github.com/ImperialCollegeLondon/ve_data_science/blob/984ed28da6e7e963bb8c34b52f72ee16b9669b02/docs/animal_model_validation_plan.md)

TaranRallings:
Plan looks good to me!

PR #464, opened 2026-07-31
455 soil litter cnp ratio1 troubleshoot

hrlai:
This PR addresses a bug in the conversion of soil and litter input data to netCDF, which rendered the CNP ratios to cluster around 1 as reported by @jacobcook1995 in #455 

## Bug 

The culprit was a missing transpose in the converter function `tools/R/R/convert_df_to_nc.R`, which is now fixed.

## Additional work

- I also added unit tests for the converter function to minimise this from recurring
- Then the other files received minor edits to polish

## How it changes the CNP ratios

Compared to the [wrong ratios](https://github.com/ImperialCollegeLondon/ve_data_science/issues/455#issuecomment-5124485398), the updated ratios now look as follow. My take away is that the CNP ranking/order looks sensible, though the actual ratios could still slip an order of magnitude smaller or larger. **Do you think this is still problematic for initialisation** @jacobcook1995 ? (I guess it settles some of the troubleshooting of yours on the programming side, but for initialisation per se it is still imperfect.)

image
image

Note that the single-bar histogram of the necromass pool is a result of a fixed ratio used across grids due to lack of data.

## Data on Globus

The soil and litter input data have been updated for both the Maliau scenarios.


jacobcook1995 (approved):
Thanks for the quick fix on this!

(also, adding testing when areas of the code cause issues is a good plan)

PR #480, opened 2026-08-04
396 replace poetry with uv in ci r tests

hrlai:
This PR ensures the automated R tests on CI pass by:
- replacing `poetry` with `uv`
- setting a proper relative path for RETICULATE_PYTHON

I also removed the pinned version of pyrealm (I forgot when or who pinned that), as it does not seem to be essential anymore.

I'm actively developing a few R tools so would like to have this passing, so I don't break anything in the process... Thanks!


arne-exe (approved):
Approving but perhaps good to wait for David/Sally approval to make sure they're on board.

davidorme (approved):
LGTM

PR #492, opened 2026-08-06
471 initial soil carbon densities are way too high

hrlai:
This PR fixes the input soil-carbon data for Maliau 1 and 2, which was too high due to a missing conversion of [%] to [kg/kg].

The single changed file pertains to Maliau 1, but also invisibly affects Maliau 2 (because I needed to re-run the same Maliau 2 script to update the subset data). 

Both input data have been uploaded to Globus, and will update the change log in #485


jacobcook1995 (approved):
LGTM!

PR #498, opened 2026-08-07
Another copilot edit

nickwctan:
This PR adds two instructions on agents.md:

1) scanning repo - local first (include uv env) then remote 

2) read limits for remote repo: priorities and reporting constraints before proposing final fixes 

@hrlai  if you dont mind looking at this to see if this could improve/remove efficiency of copilot. Feel free to comment.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -232,6 +232,88 @@ When adding new scripts without explicit location guidance, place them in the
 closest domain folder or in [tools/](tools/) if they are reusable across
 domains.
 
+## Repository scan scope and read limits
+
+### Local-first check before remote repository access
+
+Before querying or browsing any remote repository, check local repository roots
+first. At minimum, check:
+
+- the current workspace repository root for this project
+- a sibling `virtual_ecosystem` repository if it is available in the same
+  parent directory or current workspace
+- if no local source checkout is available, the installed
+  `virtual_ecosystem` package in the active uv environment (typically under
+  `.venv/lib/python*/site-packages/virtual_ecosystem/`)
+
+If a relevant local repository exists, use it as the primary source for code
+analysis and recommendations. Only query remote repositories when local content
+is unavailable in all local sources, outdated for the task, or when the task
+explicitly requires
+remote state (for example, open PR status or remote-only branches).
+
+### Scope
+
+Analyse the repository broadly before proposing final code changes that may
+affect shared logic, public interfaces, or cross-domain behaviour.
+
+For localised single-file fixes, start with the target module and nearest
+tests, then widen scope only when evidence suggests related dependencies.
+
+Include:
+
+- [analysis/](analysis/)
+- [tools/](tools/)
+- [tests/testthat/](tests/testthat/) and module-local test directories
+- [docs/](docs/)
+- [.github/workflows/](.github/workflows/)
+- root configuration files (for example [pyproject.toml](pyproject.toml),
+  [mkdocs.yml](mkdocs.yml), [air.toml](air.toml),
+  [.pre-commit-config.yaml](.pre-commit-config.yaml), and
+  [r_requirements.R](r_requirements.R))
+
+Exclude by default:
+
+- generated artefacts and build outputs
+- local virtual environments from broad scans, including
+  [.ve_data_science](.ve_data_science), [ve_data_science/](ve_data_science/),
+  [temp/](temp/), and [notebooks/ve_ds_venv/](notebooks/ve_ds_venv/)
+- exception: when a local `virtual_ecosystem` source repository is unavailable,
+  inspect only the installed `virtual_ecosystem` package path in the active uv
+  environment before using remote sources
+- vendor-style dependency directories and binary assets
+- large data directories under [data/](data/) unless directly required
+
+### Read limits for remote repository work
+
+When working with a remote repository, treat file reads as a constrained
+resource and budget them deliberately.
+
+- Prioritise high-signal files first: root config, CI workflows, target module,
+  and closest tests.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -232,6 +232,88 @@ When adding new scripts without explicit location guidance, place them in the
...
+
+Before querying or browsing any remote repository, check local repository roots
+first. At minimum, check:
+
+- the current workspace repository root for this project
+- a sibling `virtual_ecosystem` repository if it is available in the same
+  parent directory or current workspace
+- if no local source checkout is available, the installed
+  `virtual_ecosystem` package in the active uv environment (typically under
+  `.venv/lib/python*/site-packages/virtual_ecosystem/`)
```
hrlai: This is a Unix path, suggest adding Windows path as well.

AGENTS.md
```diff
@@ -232,6 +232,88 @@ When adding new scripts without explicit location guidance, place them in the
...
+Before recommending final fixes, provide:
+
+- a Files reviewed summary with counts by directory
+- a list of unread or skipped files/directories and why they were skipped
+- an explicit constraint note when a full scan was not possible
+
+### Quality gate
+
+Do not propose final fixes until the repository scan summary above is complete.
+
```
hrlai: This section mixes local and remote scans, which could confuse the agent. I have split them into two sections instead. I also prioritised the uv version of VE over the cloned repo, because uv standardises across our work. But I also understand that you modify stuff in a local repo of VE. I would still opt for scanning an installed VE codebase in the virtual environment because scanning a cloned VE repo requires Copilot to have access to `cd` into folders outside of the current working directory and this is giving it more access to each of our local drives.

What do you think?

```suggestion

## Local repository scan and context grounding

### Local-first source selection

Before querying or browsing any remote repository, check local sources first.

Prioritise sources in this order:
1. the current workspace repository root
2. the installed `virtual_ecosystem` package in the active uv environment
3. a cloned sibling `virtual_ecosystem` repository, if present

Use the local workspace repository as the primary source for this project. Use `virtual_ecosystem` only when the task needs implementation context from that codebase, such as symbols, functions, variables, or types.

Treat the installed uv package as the preferred `virtual_ecosystem` source when both the package and a cloned repository are available.

Only use remote repository access when local sources are unavailable, clearly outdated for the task, or the task explicitly requires remote state.

## Remote repository scan scope and read limits

### Scope

For shared logic, public interfaces, or cross-domain behaviour, scan broadly before proposing changes.

For localised single-file fixes, start with the target module and nearest tests, then widen scope only when evidence indicates related dependencies.

Exclude by default:
- vendor-style dependency directories and binary assets
- large data directories under `data/` unless directly required

### Remote read limits

Treat remote reads as a constrained resource.

- Prioritise high-signal files first: root config, CI workflows, target module, and closest tests.
- Exhaust local source checks before any remote reads.
- Start with a capped initial pass of 5–10 files.
- Batch directory discovery before deep file reads.
- Defer large or low-signal files unless directly relevant.
- If read limits prevent a full scan, stop and report the constraint before proposing final fixes.

### Required output

Before recommending final fixes, provide:
- a Files reviewed summary with counts by directory
- a list of unread or skipped files/directories and why they were skipped
- an explicit constraint note when a full scan was not possible

### Quality gate

Do not propose final fixes until the repository scan summary is complete.

```

AGENTS.md
```diff
@@ -232,6 +232,88 @@ When adding new scripts without explicit location guidance, place them in the
...
+- [analysis/](analysis/)
+- [tools/](tools/)
+- [tests/testthat/](tests/testthat/) and module-local test directories
+- [docs/](docs/)
+- [.github/workflows/](.github/workflows/)
+- root configuration files (for example [pyproject.toml](pyproject.toml),
+  [mkdocs.yml](mkdocs.yml), [air.toml](air.toml),
+  [.pre-commit-config.yaml](.pre-commit-config.yaml), and
+  [r_requirements.R](r_requirements.R))
```
hrlai: This list of folders are not suitable for the remote scan because they only exist in the local ve_data_science repo. You could ask it to scan the remote ve_data_science but I think it is always far better to scan locally, for what you're working on, and if you need stuff that have been merged into `main`, just fetch and merge them into your local branch. So I suggested to remove them.

AGENTS.md
```diff
@@ -232,6 +232,88 @@ When adding new scripts without explicit location guidance, place them in the
...
+
+Exclude by default:
```
hrlai: I have also removed this exclusion list because:
1. I found Copilot to be pretty good at excluding personal virtual environments and large binary/data files already, have you found it otherwise?
2. I am concerned about excluding paths like `[ve_data_science/](ve_data_science/)`, this might confuse the agent to miss our repo entirely???

hrlai (changes requested):
Thanks @nickwctan this would be useful for the follow-up LLM constant search. I have made quite a few suggestions below, please let us know what you think.

AGENTS.md
nickwctan: Thanks for the suggestions! Happy to accept them and see how it goes.

> Copilot to be pretty good at excluding personal virtual environments and large binary/data files already

Yes, I had to convert some excel files from supplementary data of a paper in order for copilot to read 

>  excluding paths like [ve_data_science/](ve_data_science/)

It was mainly to exclude the virtual environment folder but I have not tried it yet honestly.

hrlai (approved):
LGTM now. 

By the way, we're in exploration mode now, so there will be a lot of expected redundancies in these instruction files. One day it might be good to regroup and comb through them once for a housekeeping.

PR #508, opened 2026-08-13
Return of the Jedi: parsing VE codebase to build a constant database

hrlai:
Ok sorry first of all; this pun is almost warranted.

This PR revisits the `jedi` pipeline to construct a constant parameter database for the downstream LLM to ingest and search for published values.

The motivation was a very long and critical review with Copilot. The length conversation is archived [here](https://github.com/ImperialCollegeLondon/ve_data_science/blob/e4a2117280b2a86ba96d23e5b665a9b24181f4db/analysis/soil/llm/conversation-export.md). In brief, I asked for a critical review on the RAG approach, during which I discovered that the [previous `jedi`-based inference](https://github.com/ImperialCollegeLondon/ve_data_science/issues/423) can be improved. If this can be done, then there is no need for a RAG anymore. I preferred the more deterministic `jedi` analysis so I dug deeper.

First, `tools/python/src/ve_data_tools/constant_usage_tool.py` is buffed up a lot with the help of Copilot. @jacobcook1995 and @davidorme this is the file that I need the most help with. Could you please review it? Warning ahead, it is very long ~1000 lines now. But many of it are helper function; the single user-facing function `get_constant_references()` starts at Line 706. Happy to remove any fluff, I am aware that it can be too long.

From a few repeated runs and later formal use in the R script `analysis/soil/llm/extract_constant_metadata.R`, it seems to be doing the job quite well. Most importantly, the [previous reported issue](https://github.com/ImperialCollegeLondon/ve_data_science/issues/423) is now resolved: `reference_cue_logit` is now linked to `calculate_carbon_use_efficiency` correctly.

I also added a defense against the `1 / SoilConstants.x` case @davidorme suggested. This will be categorised as "needing human review". For the review or ground-truthing part, I sketched a script that generates a reviewer-friendly spreadsheet `analysis/soil/llm/generate_constant_review_workbook.R`, but this would be easier if @jacobcook1995 and I have a brief meeting to go over it --- for now it uses soil constants as a testing ground for human reviewer; we should discuss if this is necessary, and if so is the spreadsheet easy to use.

Thanks!


tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+
+Notes for R users:
+    A Python ``dict`` is similar to a named R list. A Python ``list`` is
+    similar to an unnamed R list. ``Path`` stores a file path and provides
+    methods such as ``is_file()`` and ``read_text()``.
+
+    The abstract syntax tree (AST) is similar to parsed expressions from
+    ``parse()`` in R. It describes code structure without executing code.
+    ``jedi`` then resolves names across Python files, similar to an IDE
+    "find references" workflow.
```
jacobcook1995: Minor comment, but these sorts of R vs python notes are maybe worth making a general documentation page so that there is one place to look for them (rather than having them at the top of every script)

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+    ``kwarg_forward`` / ``positional_forward``:
+        Constant is forwarded unchanged into another call. The receiving
+        callable is stored in ``consumer``.
+    ``derived_forward``:
+        A derived value (for example ``1 / constants.x``) is forwarded. The
+        consumer cannot be resolved reliably, so the site is flagged.
+    ``validator``:
+        Reference occurs inside a pydantic validator on the owning class.
+
+    Only ``derived_forward`` sites require manual review.
```
jacobcook1995: I would put this sort of information (i.e. stuff that refers to named keywords) in the docstrings of the relevant function, as you want this information to be close to the point of usage

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+    ``parse()`` in R. It describes code structure without executing code.
+    ``jedi`` then resolves names across Python files, similar to an IDE
+    "find references" workflow.
+
+Approach:
+    1. Import-based class detection: identify subclasses of
+       ``Configuration`` (including indirect inheritance via MRO).
+    2. AST-based attribute and usage analysis: enumerate declared attributes
+       and classify reference contexts from syntax.
+    3. Jedi reference resolution: find cross-file reference sites.
```
jacobcook1995: For the top level docstring I think it is better to have an extended description of why you have taken a particular approach rather than just stating the approach as a list of steps (that sort of thing is much more what you want in a function docstring)

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+    file: str
+    line: int
+    column: int
+    caller: str
+    usage_kind: str
+    consumer: str = ""
+    forwarded_as: str = ""
+    expression: str = ""
+
+    def to_dict(self) -> dict[str, Any]:
```
jacobcook1995: Is there a value in defining this method, when all it does is a standard dictionary transform?

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+    """Convert a project-relative ``.py`` path to a dotted module name.
+
+    Args:
+        relative_path: Project-relative source file path.
+
+    Returns:
+        Dotted module path without file suffix.
+
+    """
+    parts = relative_path.with_suffix("").parts
```
jacobcook1995, line 149: shouldn't this be without suffix

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+        project_root, "rev-list", "--left-right", "--count", "HEAD...@{upstream}"
+    )
+    ahead, _, behind = counts.partition("\t")
+
+    return {
+        "project_upstream": upstream,
+        "project_upstream_synced": counts != "" and ahead.strip() == "0",
+        "project_commits_ahead_of_upstream": int(ahead) if ahead.strip() else 0,
+        "project_commits_behind_upstream": int(behind.strip()) if behind.strip() else 0,
+    }
```
jacobcook1995: This function and several functions above all handle checking whether the project is upto date with develop, (doesn't have uncommitted changes etc). I'm wondering if there's much value to this, or if it's just something better handled manually?

jacobcook1995: Actually if the project provenance stuff is worth keeping, my feeling is that it should become a separate tool that gets imported by this script

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+        return "", ""
+
+    if not definitions:
+        return "", ""
+
+    definition = definitions[0]
+    return definition.full_name or "", definition.docstring() or ""
+
+
+def _first_paragraph(docstring: str) -> str:
```
jacobcook1995: function names should generally be verbs, so this should be "_extract_first_paragraph"

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+    r"""Extract summary prose from a longer docstring.
+
+    Args:
+        docstring: Raw docstring text, possibly including signature and sections.
+
+    Returns:
+        First descriptive paragraph after removing common argument/return
+        sections. Returns an empty string if no prose is available.
```
jacobcook1995: Guess another question is why only the first paragraph? If there are three descriptive paragraphs they are probably all relevant

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
-
-    project_root = Path(project_root)
+    target_file_path: str | Path | list[str | Path],
+    out_path: str | Path,
+    project_root: str | Path,
+    include_tests: bool = False,
+    progress_callback: Callable[[int, int, str], None] | None = None,
+) -> dict[str, dict[str, Any]]:
+    """Find and classify configuration constant references, then write TOML.
+
```
jacobcook1995, line 679: This is the place to add details of the function flow

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
 
-    # Collect references across all target files into a single dictionary
-    script_references = {}
+    # AST caches keyed by absolute path, so each referencing file is parsed once.
+    tree_cache: dict[Path, ast.AST] = {}
+    parent_cache: dict[Path, dict[ast.AST, ast.AST]] = {}
+    lines_cache: dict[Path, list[str]] = {}
+    script_cache: dict[Path, jedi.Script] = {}
+
+    def load(path: Path) -> tuple[ast.AST, dict[ast.AST, ast.AST], list[str]]:
```
jacobcook1995: I would avoid defining functions within a function (mainly because it's hard to read). Better to separate this out as an independent function or as a method

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
+    records: dict[str, ConstantRecord] = {}
+    functions: dict[str, str] = {}
+
+    total_targets = len(target_file_paths)
+    show_progress = total_targets > 1
+    if show_progress and progress_callback is None:
+        sys.stdout.write(f"Scanning constant usage in {total_targets} files...\n")
+        sys.stdout.flush()
+
+    def _progress_update(index: int, path: Path) -> None:
```
jacobcook1995: Same comment as above

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,975 @@
...
-        if not file_path.is_absolute():
-            full_file_path = project_root / file_path
-        else:
-            full_file_path = file_path
+        return tree_cache[path], parent_cache[path], lines_cache[path]
 
-        # Initialize a jedi Script with the source, path, and project scope
-        with open(full_file_path) as f:
-            source_code = f.read()
+    def script_for(path: Path) -> jedi.Script:
```
jacobcook1995: Same comment as above

jacobcook1995 (commented):
I've made some code quality suggestions for the python code, but generally agree that it makes most sense to discuss the approach as a whole in a meeting (when we can look through the output in detail)

tools/python/src/ve_data_tools/constant_usage_tool.py
hrlai: We discussed to relax this "first paragraph" rule. Abiotic codes contains equations in the 2nd paragraph, for example. Even `Args` are worth including.

tools/python/src/ve_data_tools/constant_usage_tool.py
hrlai: I agree that the naming is counter-intuitive, but it seems to be correct?
```python
>>> from pathlib import Path
... 
... def _module_name_from_path(relative_path: Path) -> str:
...     parts = relative_path.with_suffix("").parts
...     return ".".join(parts)
... 
... p = Path("tools/python/src/ve_data_tools/constant_usage_tool.py")
... 
... print("input:", p)
... print("with_suffix(''):", _module_name_from_path(p))
... print("without removing suffix:", ".".join(p.parts))
input: tools\python\src\ve_data_tools\constant_usage_tool.py
with_suffix(''): tools.python.src.ve_data_tools.constant_usage_tool
without removing suffix: tools.python.src.ve_data_tools.constant_usage_tool.py
```

jacobcook1995: oh `with_suffix("")` means "replace what ever suffix is there with what's in the string" (which in this case is nothing). that makes sense now

hrlai: D'oh of course. (so this is like `str_replace` in R, note to self)

hrlai:
@jacobcook1995 thanks for the chat, I incorporated your suggestion to `tools/python/src/ve_data_tools/constant_usage_tool.py`, and then split the provenance function into a standalone module in `tools/python/src/ve_data_tools/provenance.py`, plus a unit test for it.

If you could take another quick skim, that'd be great. I think it's okay to skip the other files (they are either WIP R scripts or uv files). I'm going to merge this into a working branch, so they can be reviewed at a later stage.

tools/python/src/ve_data_tools/constant_usage_tool.py
```diff
@@ -1,184 +1,903 @@
...
+imported_files:
+  - name: Configuration base class
+    path: virtual_ecosystem/core/configuration.py
+    description: |
+      Provides the ``Configuration`` superclass used to identify declaring
+      classes.
+
+package_dependencies:
+  - ast
+  - importlib
```
jacobcook1995: I don't think there's any value in listing `ast` and `importlib` are they are both part of the python standard library and weren't added there recently (`importlib` got added in 3.4, `ast` was included in python 3.0)

jacobcook1995 (approved):
Had a minor comment, but otherwise looks good to me!

tools/python/src/ve_data_tools/constant_usage_tool.py
hrlai: Thanks, I'll remove them :)

Issue #489, opened 2026-08-06
Recompile soil input data with updated soil campaign primary dataset


hrlai:
@jacobcook1995 updated the soil campaign dataset so I should recompile the input data. But wait for #471 to resolve first.

New changes:
- The soil clay, silt and sand compositions previously did not sum to 100%, these have been normalised so that this is now the case
- A new categorical variable “texture_class” has now been added to the plot data
- Instead of using “Yes/No” as categories for binary choices (“carbon_plot”, “Subsampled”), Booleans are now used instead.

PR #510, opened 2026-08-13
509 update `subset_nc()`

### LINKED ISSUE #509: Update `subset_nc()` to cope with Zarr output vs netCDF input
This might as well be a redesign, I should think about it. Having netCDF input and Zarr output has an unexpected maintenance cost of two different lines of tools...

hrlai:
This PR updates the R function `subset_nc()` to be compatible with netCDF _input_ data (again).

`subset_nc()` depends on `get_data_variables()`, but when I updated the latter to the new Zarr outputs, it lost a backward compatibility with netCDFs. 

Initially I wanted to let a single `get_data_variables()` function process both Zarr and netCDF, but it quickly becomes awkward. So I decided to just add the legacy function back as a new function `get_data_variables_nc()`. If we ever decide to also enforce Zarr _input_ data, then I can easily delete this `get_data_variables_nc()`.

This is blocking #489


nickwctan (approved):

nickwctan:
Happy to give this a pass.

PR #511, opened 2026-08-13
Variance partitioning of soil spatial model

### LINKED ISSUE #502: Variance partitioning of soil input-data predictive model
To facilitate #452 , I should provide @jacobcook1995 the variance partitioning of the soil input-data model to check if the hindcast-able predictors do explain a fair amount of variance in the soil response variables.

hrlai:
This PR calculates and visualise the variance components of the soil spatial model.

@jacobcook1995 this will help us decide how worthy hindcasting with EVI is (#452). EVI turned out not to explain a lot of the total variance across soil response variables, except maybe O-horizon depth and bulk density (EVI is the darkest-coloured bar).

image

Adding the community mean traits probably won't change the story, but we should still add them (in another PR).


jacobcook1995 (approved):
Code looks good to me so happy to approve. Only minor suggestion would be that I find the different greys a bit hard to distinguish on the graph

PR #514, opened 2026-08-14
Validation: Refactor log_dataset to screen_dataset and add_schema to improve dataset screening

hrlai:
As per title, this PR aims to streamline the user experience of dataset screening for the validation database.

## Changes

Most of the changes are in `tools/R/R/valdb.R`:
-   Add DOI normalisation and lookup
-   Store screening records under `data/derived/soil/validation/config/sources/`. This is now one YAML per dataset, instead of a huge monolithic YAML. Previously we decided for the latter, but I reverted to one per dataset, because (1) it is easier to diff, (2) easier to delete and recreate when user mistake happens.
-   Support three clearer screening decisions (`proceed`, `exclude`, and `defer`) with decision-specific reasons. No more soil-specific reasons.
-   Refactor `add_schema()` to cleanly:
    -   accept a DOI and source-record directory;
    -   require an existing `proceed` record;
    -   reject duplicate records and existing schemas

Update the validation database documentation `docs/validation_database.md`.

Retire the obsolete screening-report source without regenerating its historical HTML.

Retain the legacy aggregate `sources.yaml` as migration input because some records and schemas remain unreconciled. I will work on this in another PR.

## Validation

-   [x]  Schema tests: 48 expectations passed
-   [x]  Screening tests: 64 expectations passed
-   [x]  No failures, warnings, or skipped tests

## Notes

Existing screening records and schemas are not overwritten. Amendments require deleting the relevant monolithic YAML record and repeating the screening and schema-initialisation workflow. I am working on this in another PR #519.

Coordinate support and changes to `build_validation_database()` remain outside the scope of this PR. I will work on this in another PR.

Thanks!


hrlai:
Ah crap, I accidentally merged the second PR into this first PR... I can revert that but I'm afraid that'll cause more confusion. So, please review this PR as a combination of itself + PR #520

nickwctan (approved):
Similar goals in valdb.R compared to #522. Looks logical to me. 

Thanks for adding the screen decisions!

PR #516, opened 2026-08-16
Add script for three herbivore population-density comparisons

### LINKED ISSUE #496: Level 1 Maliau animal input check: does herbivore size or using verified- published value of density affect model behaviour?
## Why this now

Before finalising the Level 1 Maliau animal input, I want to check whether our choice of herbivore body mass representative affects the model behaviour.

I compared a large herbivore (elephant) with a small herbivore (kancil) on their own, and also with and without a supplied density. In this case, choice of animal does not matter, it's a matter of deriving their realistic adult and birth mass, hence why we need to appoint species-representative. 

The aim is to confirm:

1. whether body size and time to maturity materially change the Level 1 output;
2. whether the density input behaves as expected; and
3. whether we need to adjust the Level 1 herbivore setup before moving on.

A follow-up PR will share the notebook and test data sources used for the comparison.

Baizurah:
The script does: 

-  Compares population-density trajectories for three herbivore simulation outputs
-  Calculates summary metrics for the comparison
-  Intended to support the herbivore test workflow

Basically the same general pop-density script but I've added/modified so I can compare three outputs in one script.

### DIFF
--- a/analysis/animal/herbivore_test/fg_population_density_three_outputs_metrics.py
+++ b/analysis/animal/herbivore_test/fg_population_density_three_outputs_metrics.py
@@ -508,7 +508,7 @@ def main() -> None:
     cohort_files = {
         "elephant": data_dir / "animal_cohort_data_elephant.csv",
         "kancil": data_dir / "animal_cohort_data_kancil.csv",
-        "kancil_density": data_dir / "animal_cohort_data_kancil_density.csv",
+        "kancil_density": data_dir / "animal_cohort_data_kancildensity.csv",
     }
 
     # Grid settings shared by the three simulations.
--- a/analysis/animal/herbivore_test/fg_population_density_three_outputs_metrics.py
+++ b/analysis/animal/herbivore_test/fg_population_density_three_outputs_metrics.py
@@ -334,7 +334,6 @@ def calculate_fg_population_density(
     )
 
 
-
 def summarise_test_comparison(combined_df: pd.DataFrame) -> pd.DataFrame:
     """Summarise population-density metrics used to compare herbivore tests.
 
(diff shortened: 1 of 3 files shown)


analysis/animal/herbivore_test/fg_population_density_three_outputs_metrics.py
```diff
@@ -0,0 +1,569 @@
...
+
+    figure.tight_layout()
+
+    if output_path is not None:
+        figure.savefig(output_path, dpi=150)
+        print(f"Plot saved to {output_path}")
+
+    plt.show()
+
+
```
nickwctan, line 500: Is there any difference between the codes above this line and the codes from #241?

Since you are making a tool with multiple functions in #241 , you can import it here using:
`
from "path_to_tool" import "your_density_tool"`

With this line, you are sourcing the functions you created in the script in #241 so you dont have to type it all out again. 

For example, see Lela's[codes starting from this line](https://github.com/ImperialCollegeLondon/ve_data_science/blob/e93c092075631c62a07cc96e743cbfa7cb4ec34a/analysis/abiotic/elevation_data/maliau_elevation_data_processing_script.py#L86) 

So, in your case, it would be something like: 
```
from "path_to_tool" import "your_density_tool"

data_dir = Path("path/to/herbivore_test")

cohort_files = {
        "elephant": data_dir / "animal_cohort_data_elephant.csv",
        "kancil": data_dir / "animal_cohort_data_kancil.csv",
        "kancil_density": data_dir / "animal_cohort_data_kancil_density.csv",
    }

...(and so on with the rest of your script)
```

Remove `def main() -> None:`

nickwctan (commented):
I have not looked at the codes properly but looks good at first. My main suggestion is to remove the functions here as they are the same as #241 . It would save you alot of copy and paste and less code repetition.

Happy to talk about it further on how this works.

analysis/animal/herbivore_test/fg_population_density_three_outputs_metrics.py
Baizurah: I’ll keep it in mind once #241 is merged, as that general tool is still a work in progress. For now, I’ve included the script I used for the output so @TaranRallings can check whether I’ve done anything wrong, especially any unit mismatches, when testing the density override of for whatever reason in the findings. @nickwctan

Baizurah:
@TaranRallings

TaranRallings:
Let's pause this briefly until I've fixed the mass transfer bugs and then take another look.

TaranRallings:
Now that I've got those bugs sorted, can you rerun this and see how we're looking?

PR #519, opened 2026-08-17
Validation: switch to per-dataset schema for improved maintainability

hrlai:
This PR move the single monolithic YAML config of screened validation datasets to separate YAMLs because, after a few test runs, I still find multiple YAMLs easier to maintain, diff, and makes a PR more reviewer friendly. Most importantly, it makes it much easier for a user to add schema to each config, rather than having to scroll through a long list.

This is an independent, but companion PR to #514.


nickwctan (approved):
Following...

Issue #292, opened 2026-04-27
Document functions that are sourced from another script


hrlai:
Sometimes we source a function from another script, and this is currently not uniformly documented across our `analysis` scripts. @arne-exe left some broad ideas [here](https://github.com/ImperialCollegeLondon/ve_data_science/pull/289#pullrequestreview-4178201818).

I think the near-term actionable task is to implement a documentation for _ourselves_, and leaving the public facing documentation (package, readthedocs, visualisation flow chart etc.) for later. The former can be achieved as simply as adding a new `source_files:` section to the scripts' metadata chunk, as suggested by @davidorme [here](https://github.com/ImperialCollegeLondon/ve_data_science/pull/289#issuecomment-4325627180). If so, then a PR would target the `template` scripts.

I am thinking about this in the R context, though it might extend to Python's `import` etc. Thoughts?

nickwctan:
> I am thinking about this in the R context, though it might extend to Python's import etc. Thoughts?

Yeah I'd say same goes for python. Would it not be sufficient to add a line or two of annotations under the each function source?

hrlai:
> Would it not be sufficient to add a line or two of annotations under the each function source?

@arne-exe what's your thought?

hrlai:
Hi @nickwctan @arne-exe @lelavathy @Baizurah , closing this as this mostly concerns my existing scripts, which now mostly has a `source_files:` field in the metadata header. Please add this field whenever you source a file in your scripts.

Some of us will have some older scripts that needs backfilling, but this is absolutely not urgent, and can be very easily aided by Copilot later.

hrlai:
I should also add that the R template in `templates/R_script_template.r` already has a `source_files:` field. Not sure if Python folks want to add that to `templates/python_script_template.py`.

nickwctan:
Good idea. I'll add it in.

PR #528, opened 2026-08-18
adding imported file as metadata

nickwctan:
Our python scripts have not yet gotten a field for imported files in the metadata header. Adding this might help the user understand what the scripts do and add clarity. See #292 

Fixes #572


lelavathy:
This looks good to me. This should make the dependency easier to understand without needing to inspect the imported code.

lelavathy (approved):

PR #529, opened 2026-08-18
Updated job array submission scripts.

### LINKED ISSUE #512: Integrate sensitivity analysis scripts with PBS array job execution on CX3
Integrate the sensitivity analysis scripts (@lelavathy) with the draft HPC array job configuration (@davidorme) to standardise HPC model execution. 

## Goal:
Execute all runs defined in a config.toml file through an easy to use automated array job submission script.

### Expected User Workflow:
1. Clone the repository to a Home HPC directory. (data can be accessed by referring to project space)

2. Initialise a virtual environment using uv.

3. Generate the config.toml file (e.g., via analysis/abiotic/sensitivity/morris_sample.py). This will define the parameters to adjust, a secondary config would define the remaining default parameters. 

4. Submit the array job using a single bash script.

## Criteria:
Dynamic Array Size: The submission script should read the config.toml file and automatically set the PBS array size to match the number of job entries.

Task Mapping: Each PBS subjob to execute the Veco model using the specific parameter set defined by its corresponding [[jobs]] index. 
(bonus: include the option to add repeats)

Robustness: A subjob failure should not terminate the remaining tasks in the array.

Docs: to introduce new HPC users to the workflow.

### Unresolved:

Currently the draft has a fixed resource request. 
proposed solution: define the required resources when generating the config.toml.

#### Note 
_I will make these changes on another branch "hpc-rse-support", and open child issues to document the changes made. 
Then when it's functional I will submit a PR for your review :)._ 

_Do let me know if you have any suggestions in the meantime._

### LINKED ISSUE #513: Develop minimal functioning PBS array submission pipeline
- use the .toml produced my morris_sample.py.
- make minimal changes to the existing HPC execution draft.
    - Use the existing conda env.
    - Use a fixed resource allocation

moi-taiga:
Updates the draft for array job submission.

Changes & key functionality:
Configures submit_ve_array_job.sh to parse job_config.toml and execute an array job using the defined parameters.
Replaces the shared Conda environment with a user-specific .venv built via uv.
testing was done using the example dataset and the user version of Veco.

TODO:
HPC resource allocations (e.g., -lselect=1:ncpus=1:mem=1gb) remain hardcoded for the example dataset.
@lelavathy - we should discuss the optimal approach for defining these resource parameters within the workflow.


docs/hpc_setup.md
```diff
@@ -0,0 +1,144 @@
...
+curl -LsSf https://astral.sh/uv/install.sh | sh
+```
+
+Create the project venv and install the repository dependencies:
+
+```bash
+uv sync
+```
+
+(TODO check that uv sync is giving the correct versions)
```
hrlai: Hi @moi-taiga , we wrote a guide to set up `uv` lately here. It would be great to get your thoughts on it and/or link to it from here. It should be compatible with HPC work, but when writing it I haven't really thought about HPC (only each of our local computers). Thanks.

docs/hpc_setup.md
```diff
@@ -0,0 +1,144 @@
...
+submitting jobs. Computationally intensive work should be submitted to compute
+nodes through PBS.
+
+## Clone the repository
+
+Keep a personal checkout (like you would locally) rather than working from a
+shared project directory. This avoids conflicts with other users' files.
+
+```bash
+cd ~
```
hrlai, line 42: @moi-taiga cloning the repo here will lose the files synced with Globus to `/rds/general/project/virtual_rainforest/live/ve_data_science/` ? Sorry we should've said earlier that we have Globus set up there to transfer large files that are not pushed to GitHub. These files will be used as inputs for `ve_run`.

docs/hpc_setup.md
moi-taiga: I will have a look at your `uv` guide and see if it is compatible with the HPC workflow. :)

docs/hpc_setup.md
moi-taiga: I think it may still be best for individuals to clone the repo. You can point to the `/rds/general/project/virtual_rainforest/live/ve_data_science/` path when running a job (possibly for both input and output data if it's a big job). 

Let me know if you think that wouldn't work.

hrlai: No actually that sounds good. This should avoid virtual environment conflicts too. Then each of us simply read data from the `rds` directory. This requires us to update the file paths in our scripts (we have mostly been relativising it to the project root, but not a big deal :)

moi-taiga:
@lelavathy Could you take a look at the latest updates and see if the job_config.toml is still in a format which your sensitivity output could match? 
I could assist in modifying the sensitivity analysis scripts (morris.py etc) if needs be. 

note that job names have been removed in favour of outputs being named according to pbs ID.

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+Use SSH to access the cx3 cluster; you will be required to use your password for authentication.
+
+```bash
+ssh userid@login.cx3.hpc.imperial.ac.uk
+```
+
+See the [Imperial RCS documentation](https://icl-rcs-user-guide.readthedocs.io/en/latest/hpc/getting-started/)
+for further information about connecting to the cluster.
+If you are new to HPC environments, do make sure that you familiarise yourself with the
+documentation.
```
dalonsoa: Maybe mention the introduction to the Imperial HPC course. Something like:

```suggestion
If you are new to HPC environments, do make sure that you familiarise yourself with the
documentation. It is highly recommended to attend the [Introduction to HPC at Imperial course](https://www.imperial.ac.uk/early-career-researcher-institute/learning-and-development/courses-by-programme/research-computing-and-data-science/introduction-to-hpc/), but if you cannot make it, the materials for the course are freely available and give a good overview of how to work with the system.
```

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+- `site_directory`: the absolute path to the directory containing the shared
+    configuration files and input data. Replace the path in the example
+    with the path to your checkout's `ve_example` directory.
+- `common_config_paths`: paths to configuration files, relative to
+    `site_directory`, that every array sub-job uses.
+- `[[jobs]]`: one or more job entries. Each can provide additional
+    `config_paths` and a `config` table of values that override the loaded
+    configuration.
+- `repeats`: a positive integer on each job entry. The total number
+    of PBS array sub-jobs is the sum of all `repeats` values.
```
dalonsoa, line 85: Why is this? Is there stochasticity in the runs?

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Submit the PBS array job
+
+The submission script takes three arguments:
+
+1. the batch TOML file;
```
dalonsoa: Is this the above file you just described? If so, in the section above call it consistently `job configuration` and do not jump between `configuration` and `specification`, which might be equivalent but also might cause confusion.

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+one child directory per PBS array sub-job. You can submit from any working directory:
+
+```bash
+bash hpc_jobs/submit_ve_array_job.sh \
+    hpc_jobs/job_config.toml \
+    hpc_jobs/resources_config.toml \
+    ve_example/<experiment-output>
+```
+
+The submission script validates the job specification and resources configuration,
```
dalonsoa: Same here

```suggestion
The submission script validates the job configuration and resources configuration,
```

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+`hpc_jobs/resources_config.toml` has a small allocation suitable for the example data
+
+Every value applies to each individual array sub-job,
+apart from `max_concurrent_jobs`, which caps how many sub-jobs the scheduler runs at once.
+
+Larger datasets may require more memory, CPUs, or wall time.
+More jobs don't need more resources as resources are set per job.
+
+`hpc_jobs/resources.py` loads and validates the resource configuration, while
+`hpc_jobs/hpc_ve_job_spec.py` validates the job specification. Each resource field is
```
dalonsoa: And again here. Either `configuration` or `specification` but do not mix them.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+    def __post_init__(
+        self,
+    ) -> None:
+        """Populate the total num of jobs and map jobs to respective array indices."""
+        # Map each job array index to the corresponding job index.
+        self.job_map = [
+            job_index
+            for job_index, job in enumerate(self.jobs)
+            for _ in range(job.repeats)
+        ]
```
dalonsoa: As this sits now, each repeat will have the same index. It will look something like `[0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3]` for the case of 4 jobs with 3 repeats each, for example? 

Is this correct? And the values will go from `0` to `len(self.jobs)`

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+        # calculate the total number of jobs based on the job map.
+        self.n_jobs = len(self.job_map)
+
+    def get_job(self, array_index: int) -> Job:
+        """Get the correct job for a job array index."""
+
+        # I cant see how this would happen but just incase avoid a silent failure.
+        # ensure that the array index is within the valid range of jobs.
```
dalonsoa:
```suggestion
        # I can't see how this would happen but just incase, avoid a silent failure.
        # Ensure that the array index is within the valid range of jobs.
```

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+    repeats: int = Field(default=1, ge=1)
+
+
+@dataclass
+class JobSpec:
+    """Defines a job specification."""
+
+    common_config_paths: list[str]
+    site_directory: str
+    jobs: list[Job]
```
dalonsoa: Let pydantic do its job:

```suggestion
    jobs: list[Job] = Field(min_length=1)
```
And remove the add-hoc check done below.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+
+    """
+    with open(job_file, "rb") as jobs:
+        data = tomllib.load(jobs)
+
+    # Ensure there is at least one job specified.
+    if not data.get("jobs"):
+        raise ValueError(
+            "No jobs defined. Add at least one [[jobs]] section to the JOB_CONFIG.toml."
+        )
```
dalonsoa: Not needed if you let pydantic do what does best, which is validate inputs.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+    # Validate the job specification using Pydantic
+    # (calls the __post_init__ method to populate n_jobs and job_map)
+    job_spec = JobSpec(**data)
+
+    # Validate that the site directory exists and is a directory
+    site_directory = Path(job_spec.site_directory)
+    if not site_directory.is_dir():
+        raise ValueError(
+            f"Site directory does not exist or is not a directory: {site_directory}"
+        )
```
dalonsoa: This should go into a [model validator](https://pydantic.dev/docs/validation/dev/concepts/validators/#model-after-validator) in pydantic.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+
+    missing_paths = [
+        str(site_directory / path)
+        for path in config_paths
+        if not (site_directory / path).is_file()
+    ]
+
+    if missing_paths:
+        formatted_paths = "\n".join(f"  - {path}" for path in missing_paths)
+        raise ValueError(f"Config files do not exist:\n{formatted_paths}")
```
dalonsoa: And this as well.

hpc_jobs/examaple_hpc_configs/arrayJob_config.toml
dalonsoa, line 1: I would keep the example configuration files in a separate directory. As they are examples, mixing them with code is not good practice.

hpc_jobs/examaple_hpc_configs/pbs_resources_config.toml
dalonsoa, line 1: Same with this one

hpc_jobs/resources.py
```diff
@@ -0,0 +1,70 @@
...
+
+    with open(resources_config_file, "rb") as handle:
+        data = tomllib.load(handle)
+
+    return Resources(**data.get("resources", {}))
+
+
+if __name__ == "__main__":
+    # this should not really be invoked because the script will be run in
+    # submit_ve_array_job.sh (with its own checks)
```
dalonsoa: But this is indeed invoked by the submit_ve_array_job.sh script, right? This comment is confusing.

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+## Clone the repository
+
+Keep a personal checkout (like you would locally) rather than working from a
+shared project directory. This avoids conflicts with other users' files.
+
+```bash
+cd ~
+git clone https://github.com/ImperialCollegeLondon/ve_data_science.git
+cd ve_data_science
+```
```
dalonsoa, line 46: I really, really strongly suggest that people uses a specific version/release of the respository, and not just cloning it and use whatever is in the main branch at that specific point in time. Otherwise, tracing the reproducibility or fixing things if something goes wrong is going to be a pain.

It should be as easy as tagging commits in the repo for specific, interesting points in time, and then:


```suggestion

```bash
cd ~
git clone https://github.com/ImperialCollegeLondon/ve_data_science.git
cd ve_data_science
git checkout tags/X.Y.Z
```

```

Even better if when running this workflow, the version used is printed in the logs somewhere.

File: docs/hpc_setup.md [hunk 16]
```diff
@@ -0,0 +1,173 @@
...
+## Install `uv` and create a virtual environment
+
+See [uv setup](uv_setup.md) for instructions about using uv.
+
+**note:** If you are using R, conda may be better suited than uv.
+**note:** for test runs with the example data I use:
+
+```bash
+uv venv
+uv pip install virtual-ecosystem
```
cc-a: This will install the latest release on pypi which has no particular compatibility guarantees. Would strongly advise the virtual-ecosystem version is pinned and recorded somewhere.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def get_job(self, array_index: int) -> Job:
+        """Get the correct job for a job array index."""
+
+        # I cant see how this would happen but just incase avoid a silent failure.
+        # ensure that the array index is within the valid range of jobs.
+        if not 1 <= array_index <= self.n_jobs:
+            raise ValueError(
+                f"Job array index must be between 1 and {self.n_jobs}: {array_index}"
+            )
```
cc-a: If the main function of this is to provide a more human readable error then I'd suggest preferring a try-except around the job_map lookup below.

hpc_jobs/hpc_ve_job_spec.py
cc-a: I'd actually go the opposite way on this and decouple these checks from loading the config entirely. Checking that things exist on disk is separate from input validation. I also think it should be possible to use a function like `load_job_spec` without requiring anything to exist on disk at all.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+    config: dict[str, Any]
+    repeats: int = Field(default=1, ge=1)
+
+
+@dataclass
+class JobSpec:
+    """Defines a job specification."""
+
+    common_config_paths: list[str]
+    site_directory: str
```
cc-a: As above.

```suggestion
    common_config_paths: list[Path]
    site_directory: Path
```

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+
+from pydantic import Field, ValidationError
+from pydantic.dataclasses import dataclass
+
+
+@dataclass
+class Job:
+    """Defines an HPC VE_run job."""
+
+    config_paths: list[str]
```
cc-a: This can be directly typed as Path and will then do more checks to make sure the string provided is a valid path for the current platform.
```suggestion
    config_paths: list[Path]
```

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+@dataclass
+class JobSpec:
+    """Defines a job specification."""
+
+    common_config_paths: list[str]
+    site_directory: str
+    jobs: list[Job]
+
+    n_jobs: int = Field(init=False)
+    job_map: list[int] = Field(init=False)
```
cc-a: Another name of this would be nice as it's not immediately clear what mapping is being made.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+from pydantic import Field, ValidationError
+from pydantic.dataclasses import dataclass
+
+
+@dataclass
+class Job:
+    """Defines an HPC VE_run job."""
+
+    config_paths: list[str]
+    config: dict[str, Any]
```
cc-a: Ideally you'd be able to validate these config values as well but this would presumably depend on that data being easily available from the VEco package.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+
+        # Get the job index from the job map and return the corresponding job.
+        job_index = self.job_map[array_index - 1]
+
+        # Return the corresponding job.
+        return self.jobs[job_index]
+
+
+def load_job_spec(job_file: Path) -> JobSpec:
```
cc-a: I generally find a file-like object is a more useful interface for this sort of function. Again, tends to make testing easier without having to write files to disk.

hpc_jobs/hpc_ve_run_batch.py
```diff
@@ -1,26 +1,28 @@
...
-local_dir = os.getcwd()
-site_dir = shutil.copytree(batch_job_spec.site_directory, local_dir)
-os.chdir(site_dir)
+# the output however contains MANY files
+# so we will likley need to stage the output directory (set to $TMPDIR)
+# then compress into tarball and copy back to the final output location
+
+# in the meantime... for small jobs, work from site dir.
+os.chdir(batch_job_spec.site_directory)
+print(f"[DEBUG] cd site dir: {os.getcwd()}", flush=True)
```
cc-a: If we're getting into this sort of territory than a proper logging framework is the way to go.

hpc_jobs/hpc_ve_job_spec.py
```diff
@@ -0,0 +1,133 @@
...
+    if not site_directory.is_dir():
+        raise ValueError(
+            f"Site directory does not exist or is not a directory: {site_directory}"
+        )
+
+    # Validate that all config paths exist relative to the site directory
+    config_paths = {
+        *job_spec.common_config_paths,
+        *(path for job in job_spec.jobs for path in job.config_paths),
+    }
```
cc-a: Related to the above, isn't this checking that every jobs directories exist every time the config file is loaded? In the scenario that I'm an array sub job and loading the config to run a single job it might make sense to check the config paths for that one job but not all of them.

hpc_jobs/submit_ve_array_job.sh
cc-a, line 1: Overall I think this would be better of as a Python script as well. This greatly simplifies interacting with the other Python components, particularly `resources.py` where evaling the output to set some environment variables is quite clunky.

docs/hpc_setup.md
moi-taiga: I believe there is some.
@davidorme requested that we maintain the functionality of selecting num repeats from his original draft.

hpc_jobs/hpc_ve_run_batch.py
```diff
@@ -29,21 +31,29 @@
...
+print(
+    f"[DEBUG] output_dir: {output_dir.resolve()}  exists={output_dir.exists()}",
+    flush=True,
+)
 
 # Update the config to use that as the output directory.
-cli_config["core.data_output_options.out_path"] = out_dir
+cli_config.setdefault("core", {})
+cli_config["core"].setdefault("data_output_options", {})
+cli_config["core"]["data_output_options"]["out_path"] = str(output_dir)
```
cc-a: Slightly neater.

```suggestion
core_config = cli_config.setdefault("core", {})
output_options = core_config.setdefault("data_output_options", {})
output_options["out_path"] = str(output_dir)
```

cc-a (dismissed):
Good work @moi-taiga. My main feedback would probably be move from the bash submission wrapper to a Python one. I'd also move a lot of logic onto methods of the config class - so having methods to check directories/paths for individual jobs, for all jobs, etc, rather than having that logic spread around.

hpc_jobs/hpc_ve_job_spec.py
dalonsoa: That's is true. I'd go with Chris suggestion on this.

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+`hpc_jobs/resources_config.toml` has a small allocation suitable for the example data
+
+Every value applies to each individual array sub-job,
+apart from `max_concurrent_jobs`, which caps how many sub-jobs the scheduler runs at once.
+
+Larger datasets may require more memory, CPUs, or wall time.
+More jobs don't need more resources as resources are set per job.
+
+`hpc_jobs/resources.py` loads and validates the resource configuration, while
+`hpc_jobs/hpc_ve_job_spec.py` validates the job specification. Each resource field is
```
moi-taiga:
```suggestion
`hpc_jobs/hpc_ve_job_spec.py` validates the job configuration. Each resource field is
```

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+    `config_paths` and a `config` table of values that override the loaded
+    configuration.
+- `repeats`: a positive integer on each job entry. The total number
+    of PBS array sub-jobs is the sum of all `repeats` values.
+
+## Submit the PBS array job
+
+The submission script takes three arguments:
+
+1. the batch TOML file;
```
moi-taiga:
```suggestion
1. the job configuration TOML file;
```

docs/hpc_setup.md
```diff
@@ -0,0 +1,173 @@
...
+2. the resources TOML file (see [Choose job resources](#choose-job-resources));
```
moi-taiga:
```suggestion
2. the resources configuration TOML file (see [Choose job resources](#choose-job-resources));
```

hpc_jobs/resources.py
moi-taiga: I see the confusion. I wrote the comment to refer to the below check for 2 input arguments.

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: Yes that is correct, so that job n is repeated (repeats) number of times.

hpc_jobs/resources.py
moi-taiga:
```suggestion
```

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: I decided against this. I started to use the validation from the VEco package as a way to avoid running jobs which would ultimatlely fail. I therefore did the check on all nJobs (as they all have different input values) but this of couse takes time and is then immediatelly run again when the jobs start. 

The decision whether to stop only the jobs which fail the checks, or all jobs is also not trivial (e.g would a percentage of failing jobs mean you no longer want to run the whole array.)

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: Ah, this was an oversight. I will change it do check all of the configs paths once before running qsub.

hpc_jobs/examaple_hpc_configs/pbs_resources_config.toml
davidorme: I also think this (and the accompanying `resources.py`) should be part of `job_config.toml` and form part of the root job configuration, rather than having two different files and modules.

hpc_jobs/submit_ve_array_job.sh
davidorme: I've wondered about this before but have always run into a bootstrap problem with finding the python executable. I've always used a bash script to submit jobs because  `#! /bin/bash` just works out of the box and the script then uses `module load conda`  or `source /path/to/activate` to launch the python environment ready for the actual python script defining the job.

The submission nodes don't have a `python` executable by default, so users would either have to activate one before using the script? Alternatively they could provide the path at the top of the script as `#! /path/to/python` but because there isn't a default python, this would be user specific alterations to a shared script.

There may be a neat solution I don't know about?

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: I've decided to use the ve_run() to check all the configs once before array job submission. (but making it optional)

hpc_jobs/examaple_hpc_configs/pbs_resources_config.toml
moi-taiga: The reason they are separate at the moment is the config which defines the inputs for VE will be automatically generated by the sensitivity analysis scripts so this avoids having to manually edit the configs generated by morris_sample.py for example.

docs/hpc_setup.md
moi-taiga: I have since made sure the wording is consistent across the scripts.

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: decoupled, function now accepts file-like input rather than a path.

docs/hpc_setup.md
moi-taiga: now uses pinned version

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: function now accepts file-like input rather than a path.

hpc_jobs/hpc_ve_run_batch.py
moi-taiga: I was using these print statements while  developing, I have since removed them

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: naming has been updated to be more verbose throughout.

hpc_jobs/hpc_ve_job_spec.py
moi-taiga: U turn: I have used VEco to check the inputs before submitting jobs. by defult the check runs and stops submission if any fail. 
It is optional so a user can force an array job to be submitted even if >0 are known to fail.

hpc_jobs/submit_ve_array_job.sh
moi-taiga: I have used `subprocess.run(qsub_command, input=pbs_script,)` which uses the active interpreter.
running `uv run --group hpc-pinned python -m hpc_jobs.submit_ve_array_job` uses an environment which works with the example.

docs/hpc_setup.md
```diff
@@ -0,0 +1,215 @@
...
+git clone --branch hpc-workflow-preview-v1 --single-branch \
+    https://github.com/ImperialCollegeLondon/ve_data_science.git
+cd ve_data_science
+```
+
+## Install `uv` and create a virtual environment
+
+See [uv setup](uv_setup.md) for instructions about using uv.
+
+A pinned verion of virtula ecosystem is available for running the example:
```
dalonsoa:
```suggestion
A pinned version of Virtual Ecosystem is available for running the example:
```

hpc_jobs/parse_arrayJob_config.py
```diff
@@ -0,0 +1,73 @@
...
+            for _ in range(job.repeats)
+        ]
+        # calculate the total number of subJobs (inc repeats) based on the job map.
+        self.n_subJobs = len(self.subjob_repeats_map)
+
+    def get_subJob(self, array_index: int) -> SubJob:
+        """Get the correct job for a job array index."""
+
+        # Get the job index from the job map and return the corresponding job.
+        job_index = self.subjob_repeats_map[array_index - 1]
```
dalonsoa, line 60: Might be worth mentioning in the docstring that PBS array indices start in 1, and hence the `-1`.

hpc_jobs/parse_arrayJob_config.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        self.subjob_repeats_map = [
+            job_index
+            for job_index, job in enumerate(self.subJobs)
+            for _ in range(job.repeats)
+        ]
+        # calculate the total number of subJobs (inc repeats) based on the job map.
+        self.n_subJobs = len(self.subjob_repeats_map)
+
+    def get_subJob(self, array_index: int) -> SubJob:
```
dalonsoa, line 48: You should consistently document the input and output arguments, as you do in other places.

hpc_jobs/examaple_hpc_configs/pbs_resources_config.toml
dalonsoa: I also think it makes sense to keep them separated, in this case.

hpc_jobs/run_subJob.py
```diff
@@ -0,0 +1,49 @@
...
+from pathlib import Path
+
+from virtual_ecosystem.main import ve_run
+
+from hpc_jobs.parse_arrayJob_config import load_arrayJob_spec
+
+# Get the command line arguments
+batch_file = Path(sys.argv[1])
+pbs_array_index = int(sys.argv[2])
+output_dir = Path(sys.argv[3])
```
dalonsoa: I know this script is not meant to be used directly by users, but it could. And it that case, I think it is way nicer to use a proper input argument parser. This bit here will become (adapt text as needed):

```python
import argparse
import sys
from pathlib import Path


def parse_args():
    parser = argparse.ArgumentParser(
        description="Process a batch file for a given PBS array index and write results to an output directory."
    )
    parser.add_argument(
        "batch_file",
        type=Path,
        help="Path to the batch file to process."
    )
    parser.add_argument(
        "pbs_array_index",
        type=int,
        help="PBS array job index (integer)."
    )
    parser.add_argument(
        "output_dir",
        type=Path,
        help="Directory where output should be written."
    )
    return parser.parse_args()


args = parse_args()

batch_file = args.batch_file
pbs_array_index = args.pbs_array_index
output_dir = args.output_dir
```

hpc_jobs/submit_ve_array_job.py
```diff
@@ -0,0 +1,181 @@
...
+                    cli_config=subJob.cli_config,
+                    validate_only=True,
+                )
+            except ConfigurationError as error:
+                raise ValueError(
+                    f"Invalid Virtual Ecosystem config for subJob: {subJob_index} \n"
+                    f"{error}"
+                ) from error
+            finally:
+                LOGGER.disabled = False
```
dalonsoa, line 107: I'd put the validation bit in a separate function. It makes this `main` function and the validation itself easier to read.

hpc_jobs/submit_ve_array_job.py
```diff
@@ -0,0 +1,181 @@
...
+        "oe",
+        "-o",
+        str(output_directory / "array_subJob_^array_index^" / "pbs.log"),
+        "-v",
+        (
+            f"VE_BATCH={arrayJob_config},"
+            f"RUN_OUTPUT_DIR={output_directory},"
+            f"ROOT_DIRECTORY={root_directory}"
+        ),
+    ]
```
dalonsoa, line 163: This feels a bit redundant - or maybe I am missing something. You are creating a simple input file programatically, and then creating a rather complicated command passing lots of arguments to that simple file. Why not creating the input file with all the information already there? You can even have the `pbs_script` available as a template in a file and just load it and fill the gaps.

dalonsoa: Something like:

### pbs_script_template.sh
```bash
#!/bin/bash
#PBS -J 1-{n_subJobs}%{max_concurrent_jobs}
#PBS -l select={select}
#PBS -l walltime={walltime}
#PBS -j oe
#PBS -o {output_directory}/array_subJob_^array_index^/pbs.log

set -euo pipefail

JOB_OUTPUT_DIR="{output_directory}/array_subJob_$PBS_ARRAY_INDEX"

cd "{root_directory}"

{python_executable} -m hpc_jobs.run_subJob \
    "{arrayJob_config}" "$PBS_ARRAY_INDEX" "$JOB_OUTPUT_DIR"
```

### Loading bit
```python
template_path = Path(__file__).parent / "pbs_script_template.sh"
template = template_path.read_text()
pbs_script = template.format(
    n_subJobs=arrayJob_spec.n_subJobs,
    max_concurrent_jobs=resources_spec.max_concurrent_jobs,
    select=resources_spec.select,
    walltime=resources_spec.walltime,
    output_directory=output_directory,
    arrayJob_config=arrayJob_config,
    root_directory=root_directory,
    python_executable=sys.executable,
)
```

tools/python/src/ve_data_tools/hpc_arrayJob_config_tools.py
```diff
@@ -122,26 +120,34 @@ def generate_job_config(
...
-    jobs = []
-    for run_id, values in enumerate(samples, start=1):
-        job = {
-            "name": f"run_{run_id:04d}",
+    subJobs = []
+    for values in samples:
+        cli_config = {}
+
+        for parameter, value in zip(parameter_names, values):
+            config_section = cli_config
```
dalonsoa, line 128: I don't think you need this. Can use directly `cli_config` and it will do the same.

dalonsoa (approved):
I have a few nitpick comments here and there, but no blockers, so approving from my side. Good work!

tools/python/src/ve_data_tools/hpc_arrayJob_config_tools.py
moi-taiga: I think you mean remove `config_section = cli_config` and then use:
`config_section = cli_config.setdefault(key, {})`
instead of 
`config_section = config_section.setdefault(key, {})`

This makes the ouput:
[subJobs.cli_config.hydrology]
[subJobs.cli_config.constants]
instead of:
[subJobs.cli_config.hydrology.constants]

dalonsoa: ah, OK. I see.

moi-taiga:
This should be ready to merge now. :) 

Zarr output is generated on the TMPDIR on the compute node, then only the NetCDF is moved to the selected output dir. 
@lelavathy see if this works for your large-scale runs.

PR #538, opened 2026-08-21
489 recompile soil input data

### LINKED ISSUE #489: Recompile soil input data with updated soil campaign primary dataset

hrlai:
This PR is now unblocked by #510. It recompiles the soil input data for Maliau using a slightly newer soil campaign data from @jacobcook1995 . During the process I also decided to log-transform the predictors before model fitting so they are less skewed.

The new soil input data have been uploaded to Globus. Change logs reflect the revision.


arne-exe (approved):
LGTM

PR #543, opened 2026-08-21
added dev only pft_name to plant definitions file

arne-exe:
Added temporary plant definitions file to maliau_1 and maliau_2 so that it can be used with the latest dev version.


hrlai (approved):
Thanks for addressing this so quickly.

Were you waiting for this to be in the VE release version? Sorry if this change request was premature.

